PR #5, opened 2015-10-06
Include some of the tools from Kadira

arunoda:
This PR is just for the discussion and not need for merging.

We've some great experience with performance monitoring of Meteor apps (and node apps). Also integration with some third party services like Librato Metrics.

So, we can specially work this chapter. Since I'd like to write stuff on this matter.
So, this the question?

* Is it okay to mention tools from Kadira? Like Kadira Profiler and FlameGraph viewer. 

> Even though, I run Kadira, I am happy to show how to integrate with other platforms. 

Looking forward for the feedback.

### DIFF
--- a/outlines.md
+++ b/outlines.md
@@ -160,8 +160,11 @@ Write “HTML with holes” just like you're used to, and get a fast, fine-grain
 Now that you've built a sweet app, give it to the world. It can be hard to run a production app, that's where Galaxy comes in.
 
 1. Integrating with popular analytics platforms to track method calls, publications, and URL hits
+2.  integrating with Librato Metrics, etc. (or just add Kadira)
 2. Profiling and monitoring your app to find out where it is slow
+3.  way to use Kadira Profiler (we'll have memory leak debuggin soon as well)
 3. Escape hatches for performance issues
+4.  Uses of Kadira Debug
 4. Staging and testing
 5. Rolling updates with hot code push
 6. Galaxy


tmeasday:
I think it's a good question. Perhaps we should rephrase as "what's our policy around commercial offerings?"

I think we'd be mad to not mention Kadira of course. But there is a spectrum of stuff with some spammy stuff at the other end (if you know what I mean). A written down policy would make it way easier than having to judge things on a case-by-case basis.

arunoda:
Yeah. I get it. I don't want to over promote Kadira. For an example, if we wanna integrate with other analytics platforms, then we can mention Kadira. (since, we are mentioning production platforms here)

I think, it's harder to write a policy for this. Since we can PR's that'd be okay.

tmeasday:
I didn't actually mean that you'd be spammy about Kadira, just that there's a spectrum of from "totally useful and everyone wants to know about it [e.g. Kadira] .. to .. "basically just spam" [some person's new consulting shop, say].

arunoda:
That's okay. Sometimes I'm spammy too :) 
Shall I work on a more in depth outline for this?

gabrielhpugliese:
I have a question that goes the way opposite of Kadira: How can I actually debug things locally without the website but with same capabilities? I like to understand how could I do on my own :)

stubailo:
After some conversation with @rdickert, I think it makes sense to put Kadira in there as it's currently the best tool for doing a lot of this stuff in a deployed Meteor app. It's not about promoting or not promoting commercial tools, it's about telling people about the best way to do something, and representing today's reality.

arunoda:
@gabrielhpugliese Basically, you need to instrument Meteor internals. Since it's JS it's not that hard to dive deep. But need care.

You can look at Kadira source code on how we do them. And also look at the [MeteorX](https://github.com/meteorhacks/meteorx) package.

gabrielhpugliese:
Oh cool didn't know it was OSS. Thanks

stubailo:
Oops - can you reopen this against the `master` branch?

arunoda:
My intention was to open a discussion and raise this point. Anyway, we got answers.
Do we need to open a issue/PR again?

stubailo:
No, probably not! Let me go and add it to the outline as a record.

arunoda:
Okay.

stubailo:
Here: 6a81f6280efa6a3145ef52d08380bc993d7abeae

arunoda:
:+1:

PR #17, opened 2015-10-08
Some Minor Changes for the Routing Guide

arunoda:
Hi, 

Just like my previous PR. This is just for the discussion purpose. Once we agree on things, we can work on real changes.

I'll start adding comments for some places as well.

### DIFF
--- a/content/routing.md
+++ b/content/routing.md
@@ -32,7 +32,7 @@ meteor add kadira:flow-router
 
 *Snippet: Adding the Flow Router package*
 
-At the time of writing this guide, Flow Router is at version 2.x.
+At the time of writing this guide, [Flow Router](https://github.com/kadirahq/flow-router) is at version 2.x.
 
 *Community packages: Flow Router is one of several popular routing packages for Meteor. Another is iron:router. You can search for router on Atmosphere to find more. Hopefully, the concepts in this routing guide will be relevant no matter which router you use, as long as it provides basic functions for URL management.*
 
@@ -74,6 +74,8 @@ The above code snippet will match certain URLs. You may notice that one of the s
 *Table: Example URLs and the resulting parameters*
 
 Note that all of the values in `pathParams` and `queryParams` are always strings since URLs don't have any way of encoding data types. You might need to use `parseInt(value, 10)` to convert strings into numbers.
+
+> FlowRouter and IronRouter internally uses a path matching library called [path-to-regexp](https://github.com/pillarjs/path-to-regexp). Check that project for advance URL pattern matching.
 
 ## Displaying different views based on the URL and defining layouts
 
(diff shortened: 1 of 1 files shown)


content/routing.md
```diff
@@ -32,7 +32,7 @@ meteor add kadira:flow-router
 
 *Snippet: Adding the Flow Router package*
 
-At the time of writing this guide, Flow Router is at version 2.x.
+At the time of writing this guide, [Flow Router](https://github.com/kadirahq/flow-router) is at version 2.x.
```
arunoda: I think it's a good idea to link to original project, when we mention them in the guide.

So, here's the idea. I am think Meteor Guide is the place for show the direction. But it should not the place for all the content. It's technically possible to do it. 

And we can't mention all the features in all these projects here. I hope it's better to mention following types of links:

* Links to the original project 
* Some discussion on the Meteor Forum - Someone might wanna know the reason for removing auth from the router. We already discuss in on the Meteor Forum. So, we can link that.
* Some articles on Wikipedia and other

content/routing.md
```diff
@@ -322,13 +329,15 @@ If some data in your app has been moved, you probably want to redirect people to
 
 As you maintain and develop new features for your app, you might discover that you need to change your URL structure. If there are already lots of links and bookmarks to your app floating around in the wild, it might be a good idea to redirect the old URLs for backwards compatibility.
 
+// XXX: This can be both done in the server side and client. In FR 3.0, we'll make this isomorphic.
```
arunoda: Are we going to mention stuff that's going to change in the future?

content/routing.md
stubailo: I think you're 100% right on this point - the guide can be the entry point for discovering these patterns, but shouldn't be the location for all of the reference material.

1. Links to original project: definitely should have this
2. Discussions about why we removed auth from the router - I don't think the guide is for historical discussions like this; I'd rather stick to guiding materials
3. Articles from Wikipedia might be relevant when talking about complicated concepts that we can't explain inline. Are there any in the routing guide?

content/routing.md
stubailo: I feel like it could just make things confusing? If people want to learn about the future of Flow Router, they can learn about it on the Flow Router website. When the new version is released we can update the guide. What do you think?

content/routing.md
stubailo: OK I've added links for all of the original projects now.

arunoda: Sound good to me. 

2. Here's the reason why I asked this: https://github.com/kadirahq/flow-router/issues/344
3. Not sure, just an idea I got.

content/routing.md
arunoda: I think that's fair and works well.

content/routing.md
stubailo: I feel like that issue thread is probably the right place for people to have that discussion, no? I don't think it's possible to defend each design decision up front. Alternatively, if you put together an article about Flow Router's design decisions, I can link to it here.

arunoda: Okay. Sounds good to me.

stubailo: Let me know if you want to write such an article or if one exists. In the meantime, going to close this PR.

PR #23, opened 2015-10-09
Add essential packages to Outline.md

mikeLspohn:
Personally, I've had to look in various places to find common packages that are used in most Meteor projects. I think a section outlining the most essential packages and where to find them would be helpful to people just coming to Meteor.

### DIFF
--- a/outlines.md
+++ b/outlines.md
@@ -104,6 +104,15 @@ What do URLs mean in a mobile and client-rendering world, and how does one use t
 5. How to handle URLs in your app that should only be accessible to certain users
 6. How and why to use a UI framework native router, like Angular router or React Router
 
+### Meteor guide: Essential packages
+
+What are some of the essential packages that most Meteor projects will need?
+
+1. Iron router - Handle routing in your Meteor application
+2. Moment - Moment.js integration for your Meteor application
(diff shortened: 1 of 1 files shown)


stubailo:
All of these will be mentioned in other parts of the guide. I don't want to structure it by "here's a cool tech you might want to use" - it's more like "I want to have user accounts in my app, what do I need to know?"

1. Routing - we are recommending flow router, there's a rough draft of a routing guide in another issue
2. Moment - there will probably be something about displaying dates correctly, but really this is just  library; I don't think is a crucial part of the app building experience.
3. NPM - that should definitely be in there 100%!
4. Accounts UI - we'll probably mention accounts-ui for prototyping, then recommend [useraccounts](https://atmospherejs.com/useraccounts/core) for more serious projects.

stubailo:
Thanks for the suggestions, btw!

mikeLspohn:
Ahhh ok, cool! That makes sense. I didn't know about flow router or useraccounts, thanks for the tip!

stubailo:
> I didn't know about flow router or useraccounts

One of the main reasons we're writing this thing! So that people can be aware of cool packages and Meteor features.

PR #57, opened 2015-10-27
Add examples of Template-Based Authorization to Further Reading

alanning:
Two examples of template-based authorization using FlowRouter + Blaze.

### DIFF
--- a/content/routing.md
+++ b/content/routing.md
@@ -441,6 +441,8 @@ That's it! The analytics package hooks into Flow Router and records all of the p
 
 Learn about advanced features of Flow Router:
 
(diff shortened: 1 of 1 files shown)


alanning:
Slides from my presentation at NY DevShop on FlowRouter + Auth can be found here:  http://slides.com/alanning55/a-pattern-for-flowrouter-auth#/

We'll also have video of the talk but not sure when that will be ready.

stubailo:
Thanks!

stubailo:
For context, this is probably not the final form of the article, but we'll make sure to add these links to the final version.

PR #229, opened 2016-02-11
Fixed initialization calls for ValidatedMethod

hwillson:
Hi guys - it looks like there are a few typos with regards to ValidatedMethod calls (using `new Method()` instead of `new ValidateMethod()`). I've updated these occurrences. Thanks!


stubailo:
Oops, this is from before we changed the name! Thanks!

PR #230, opened 2016-02-11
Changed ValidatedMethod input schema properties to validate.

hwillson:
Hi guys - I just noticed one more small thing. Some of the existing `ValidatedMethod` examples are using a `schema` property on input, but haven't defined a mixin that allows the `schema` property. To help reduce confusion I've switched these over to use `validate`. Thanks!


stubailo:
Oh gosh - it's really hard to keep these code snippets up to date, huh?

I wonder if there are any strategies for this other than going through them all manually every once in a while. Maybe PRs are the answer! Thanks!

PR #236, opened 2016-02-12
1.3 migration guide

tmeasday:
This is a list of the non-obvious imports that you'd do.

### DIFF
--- a/content/1.3-migration.md
+++ b/content/1.3-migration.md
@@ -26,28 +26,24 @@ These are all the *breaking changes* -- that is changes that you absolutely have
 
 - Cordova has been upgraded to the most recent versions (Cordova 6.0.0, Cordova iOS 4.0.1 and Cordova Android 5.1.0). This may require you to upgrade your plugin versions. We pin core Cordova plugins to versions known to be compatible and warn about this during build, but you may encounter compile time or runtime errors with third party plugins. Upgrading to newer versions of these plugins may help if they have been updated to work with recent versions of Cordova.
 
-- Some low resolution app icon and launch images sizes for now unsupported devices have been deprecated. To avoid a deprecation warning during build, please remove the entries from your `mobile-config.js`. (You probably also want to remove the image files from your project.)
-
 - When opening the generated project with `meteor run ios-device`, Xcode will show a dialog asking whether to 'Convert to Latest Swift Syntax'. Make sure to click 'cancel' here, because this conversion will leave the code in an invalid state (this seems to be an Xcode bug).
 
 - The plugin used to serve your app's files and support hot code push has been completely rewritten. As a result, files are now served from `localhost` instead of `meteor.local`, with a fixed port number derived from your `appId`. You may have to update OAuth redirect URLs to point to the new local domain and port.
-
-- The plugin now allows for local file access on both iOS and Android. You can construct file system URLs manually (`http://localhost:<port>/local-filesystem/<path>`) or use `WebAppLocalServer.localFileSystemUrl()` to convert a `file://` URL. On iOS, core plugins like `cordova-plugin-file` and `cordova-plugin-camera` (using `Camera.DestinationType.FILE_URI`) will now automatically return these URLs instead of `file://` URLs.
 
 <h2 id="modules">Recommended changes: use modules</h2>
 
 The biggest new feature in Meteor 1.3 is support for [ES2015 modules](https://developer.mozilla.org/en/docs/web/javascript/reference/statements/import) on the client and the server. Using modules you can declare dependencies between files, control load order, and use NPM packages on the client and server easily.
 
-You should load all Meteor "psuedo-globals" using the `import { Name } from 'meteor/package' syntax. For instance:
+- You should load all Meteor "psuedo-globals" using the `import { Name } from 'meteor/package' syntax. For instance:
 
 ```js
 import { Meteor } from 'meteor/meteor';
 import { EJSON } from 'meteor/ejson';
 ```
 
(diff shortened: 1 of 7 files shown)


stubailo:
We should emphasize the old way still works!

content/1.3-migration.md
```diff
@@ -0,0 +1,73 @@
+## Migrating to use imports everywhere
```
stubailo: Add title so it shows up in the TOC in the compiled version

content/1.3-migration.md
```diff
@@ -0,0 +1,73 @@
...
+
+Never use a symbol without importing it (i.e. no more psuedo-globals!).
```
stubailo: "While the old way of using Meteor core packages work, we now recommend..."

content/1.3-migration.md
```diff
@@ -0,0 +1,73 @@
...
+
+In general from core Meteor libraries your code should remain unchanged, and you just add some lines like:
+
+```js
+import {Meteor} from 'meteor/meteor';
+import {ReactiveVar} from 'meteor/reactive-var';
+```
+
+With the following exceptions
```
stubailo: Why are these exceptions? Maybe something like "These are the places where packages used to attach stuff to a different namespace, so we created a new one so you know which package it's coming from" or something.

Speaking of which, we should remind package authors to start doing this as well, like publish-composite

content/1.3-migration.md
```diff
@@ -0,0 +1,73 @@
...
+
+4. Use `import {connect} from 'meteor/ddp-client'` rather than `DDP.connect`;
+
+5. Use `import {Accounts} from 'meteor/accounts-base'` and (on the client) use:
+  1. `Accounts.user()` instead of `Meteor.currentUser()`
+  2. `Accounts.userId()` instead of `Meteor.userId()`,
+  3. `Accounts.loggingIn()` insead of `Meteor.loggingIn()`,
+  4. `Accounts.logout()` instead of `Meteor.logout()`,
+  5. `Accounts.logoutOtherClients()` instead of `Meteor.logoutOtherClient()`
+  6. `Accounts.users` instead of `Meteor.users`
```
stubailo: This looks like a case of maybe we should export `Users` directly, so it looks more like a normal collection? Not sure, and not sure if this would break some stubbing packages

content/1.3-migration.md
```diff
@@ -0,0 +1,73 @@
...
+    1. `AccountsPassword.onResetPasswordLink()`,
+    1. `AccountsPassword.addEmail()`,
+    1. `AccountsPassword.removeEmail()`,
+    1. `AccountsPassword.sendEnrollmentEmail()`,
+    1. `AccountsPassword.sendResetPasswordEmail()`,
+    1. `AccountsPassword.sendVerificationEmail()`,
+    1. `AccountsPassword.setPassword()`,
+    1. `AccountsPassword.setUsername()`,
+    1. `AccountsPassword.createUser()`,
+    1. `AccountsPassword.emailTemplate()`
```
stubailo: Are you referring to this? http://docs.meteor.com/#/full/accounts_emailtemplates

It's not a function

content/1.3-migration.md
```diff
@@ -0,0 +1,73 @@
...
+
+7. Use `import {AccountsFacebook} from 'meteor/accounts-facebook'` and:
+  1. `AccountsFacebook.loginWithFacebook()` rather than `Meteor.loginWithFacebook()`.
+
+  XXX: not yet changed on the branch, in lieu of bikeshedding
+
+8. Use `import {Template} from 'meteor/templating'` (gotcha)
```
stubailo: Oh whoa, this is interesting - cause now we have the `blaze-html-templates` package instead of `templating`... maybe as long as we document it's fine.

content/1.3-migration.md
```diff
@@ -0,0 +1,73 @@
...
+
+9. Use `import {config} from 'meteor/accounts-ui-unstyled'`
```
stubailo: Given all of the above, why not `import { AccountsUI } ...`?

content/1.3-migration.md
```diff
@@ -0,0 +1,73 @@
...
+
+10. Use `import {Computation, Dependency} from 'meteor/tracker'`
```
stubailo: `Computation` is a private constructor, it errors if you call it instead of `Tracker.autorun`. So maybe we shouldn't export it?

content/1.3-migration.md
tmeasday: Well the users collection is attached to the singleton known as `Accounts` right? I'm not sure what's best. I'd prefer it was called `Users` for sure.

I think you have to stub the properties of the collection rather than the collection itself if people are getting it via `import` anyway, no?

content/1.3-migration.md
tmeasday: Yup, this was an error

content/1.3-migration.md
tmeasday: Really? I did this on master. Do you mean that users have to `meteor add blaze-html-templates` but then import from `templating`? This is where `api.imply` makes this whole thing a bit of a mess.

tmeasday: Maybe `api.imply` should mean that you are able to `import` implied symbols from the implying package?

content/1.3-migration.md
tmeasday: Yeah, you're probably right. Will change.

content/1.3-migration.md
tmeasday: Oh it is? I didn't even realise. Will remove

content/1.3-migration.md
stubailo: Hard to say. That breaks the 1-to-1 mapping of package name to import which I would so love to preserve. I wonder if @benjamn has an opinion here.

content/1.3-migration.md
stubailo: Up to you. `Accounts.users` is fine with me as a first change.

stubailo:
I'd suggest changing "recommended changes" to "recommended change"

PR #239, opened 2016-02-16
First pass at i18n content

tmeasday:


stubailo:
I feel like actually switching languages is pretty crucial to document - since you can define translations all day but they don't do anything unless you set the right locale.

stubailo:
Other than that, looks perfect!

tmeasday:
That's true, I'll add a section

tmeasday:
Not sure the section is great (maybe we should add a language chooser to Todos?)

stubailo:
> maybe we should add a language chooser to Todos?

Sounds like we have to, in order to know what we are talking about!

tmeasday:
I suspect we won't know what we are talking about anyway; i18n is a complex thing and as english-speaking devs we (I assume in your case) have managed to mostly avoid it. I've done one tiny app with support for 2 languages and feel like I just scratched the surface..

stubailo:
Maybe we could ask the community to add it via PR?

tmeasday:
I'm not sure PRs to Todos will work awfully well, it's sort of a complete app with styles and everything. Probably we should get @dominicnguyen to add some UI for a language chooser (it'd be good anyway as it's a good example of a Store we can use)

tmeasday:
Ok, chatted to @zol about this and we aren't going to do this for 1.3, so I'd suggest we just merge this as is?

stubailo:
Sounds good, go for it!

PR #242, opened 2016-02-16
First pass at using packages

tmeasday:


outlines/using-packages.md
```diff
@@ -0,0 +1,34 @@
+# Using packages
+
+1. Packages vs modules
```
stubailo: I feel like packages vs. modules doesn't align with the actual content of this section, which is more like "when to use packages"

outlines/using-packages.md
```diff
@@ -0,0 +1,34 @@
...
+  - Packages are useful for sharing content between apps and users
+  - It's usually better to leverage someone else's work and use a package if it does what you need
+2. Npm vs Atmosphere
+  1. Npm packages:
```
stubailo: I feel like these could be the first sections of the "Atmosphere" and "NPM" sections below? Unless there's a feature table or something.

outlines/using-packages.md
```diff
@@ -0,0 +1,34 @@
...
+    - Packages can be run through the build tool (e.g. coffeescript, less packages)
+    - Packages can define build plugins!
+    - Packages are namespaced to the author
+    - Binary stuff?
+    - You can find them atmospherejs.com
+3. Installing and using Atmosphere packages
+  1. Install with `meteor add X:Y`
+  2. Installed to `.meteor/packages` -- resolved in `.meteor/versions`
+  3. `.meteor/versions` means repeatable builds (?)
+  4. Atmosphere packages can include npm dependencies, managed for you.
```
stubailo: Is this actually true, since the recommended way to do this is to ask the app developer to install it now?

outlines/using-packages.md
```diff
@@ -0,0 +1,34 @@
...
+    - The Meteor package system has some advantages
+    - Code can be targeted at client *or* server
+    - Packages can be run through the build tool (e.g. coffeescript, less packages)
+    - Packages can define build plugins!
+    - Packages are namespaced to the author
+    - Binary stuff?
+    - You can find them atmospherejs.com
+3. Installing and using Atmosphere packages
+  1. Install with `meteor add X:Y`
+  2. Installed to `.meteor/packages` -- resolved in `.meteor/versions`
```
stubailo: We should mention that the actual package files are stored in `~/.meteor/packages` and _are not_ the actual source of the package.

stubailo:
We need to forklift in the callback stuff from here: http://guide.meteor.com/build-tool.html#npm-callbacks... is `async`/`await` an option?

Also, this outline is missing some stuff from here: http://guide.meteor.com/build-tool.html#atmosphere In particular, overriding packages with local versions.

outlines/using-packages.md
tmeasday: I guess I thought I might detail why we have both and why you might choose one or the other before getting into brass tacks

outlines/using-packages.md
tmeasday: This is a subtle point. I don't think its the *recommended* way. Especially not for server-side dependencies.

outlines/using-packages.md
tmeasday: I just thought it might be useful for people to realise there's still a need to use packages even in a module world?

outlines/using-packages.md
stubailo: Fair, I guess the outline just doesn't do it justice.

outlines/using-packages.md
stubailo: I don't really see how modules would replace packages - in my mind the primary function of a package is to get other people's code, and modules don't really do that at all? IDK.

outlines/using-packages.md
stubailo: OK, definitely worth putting in then.

outlines/using-packages.md
tmeasday: OK, I'll get rid of it

tmeasday: :+1:

outlines/using-packages.md
tmeasday: :+1:

outlines/using-packages.md
tmeasday: :+1:

tmeasday:
Added that extra stuff and pushed a new version

stubailo:
Looks good! When we get to writing the content we can see how it is most natural to lay out the sections. But I think this is a complete list.

PR #244, opened 2016-02-23
Testing modules content

tmeasday:
This isn't ready yet but the PR is helpful.


content/testing.md
```diff
@@ -0,0 +1,224 @@
...
+}
+
+export default Todos = new TodosCollection('Todos');
+```
+[`imports/todos/Todos.js`]
+
+We should test that the `Todos` collection indeed behaves as we expect and sets that field when we insert a doc. To do that, we can write a file `imports/todos/todos.tests.js`, and define a Mocha test in it:
+
+```js
+import { mocha } from "avital:mocha";
```
Menda: This is wrong, `avital:mocha` does not import mocha. See: https://github.com/avital/meteor-mocha-for-1.3/issues/1

content/testing.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Defining a Mocha test
+
+In the Todos example app, we have a special `TodosCollection` that sets a `createdAt` field whenever we insert a new todo item. 
+
+```js
+class TodosCollection extends Mongo.Collection {
+  insert(doc, callback) {
+    doc.createdAt = doc.createdAt || new Date();
+    return super(doc, callback);
```
Menda: This doesn't work. Should be `return super.insert(doc, callback);`. See https://github.com/meteor/meteor/issues/6162

content/testing.md
```diff
@@ -0,0 +1,224 @@
...
+
+To run the unit tests that our app defines, we can run a special instance of our app in unit test mode. To do so, we run:
+
+```
+meteor test-app --unit --driver-package avital:mocha
+```
+
+This runs a special version of our application that:
+
+ 1. *Doesn't* eagerly load *any* of our application code as Meteor normally would.
```
Menda: I would add also: `, it loads all the dependencies imported inside tests`.

content/testing.md
```diff
@@ -0,0 +1,256 @@
...
+
+<h3 id="testing-concepts">Testing concepts</h3>
+
+Entire books have been written on the subject of testing, so we will simply touch on some basics of testing here. The important thing to consider when writing a test is what part of the application you are trying to test, and how you are verifying the behaviour works.
+
+If you are testing one small module of your application, you are writing a *unit test*. You'll need to take steps to *stub* and *mock* other modules that your module usually leverages in order to *isolate* each test. You'll typically also need to *spy* on actions that the module takes to verify that they occur.
+
+If you are testing that multiple modules behave properly in concert, you are writing an *integration test*. Such tests are much more complex and may require running code both on the client and on the server to verify that communication across that divide is working as expected. Typically an integration test will still isolate a part of the entire application and directly verify results in code.
```
stubailo: Much more complex than what?

content/testing.md
```diff
@@ -0,0 +1,256 @@
...
+
+Automated testing allows you to do all of these things to a much greater degree and *run tests more often*, which means your codebase will remain in better shape and regress less. 
+
+<h3 id="testing-concepts">Testing concepts</h3>
+
+Entire books have been written on the subject of testing, so we will simply touch on some basics of testing here. The important thing to consider when writing a test is what part of the application you are trying to test, and how you are verifying the behaviour works.
+
+If you are testing one small module of your application, you are writing a *unit test*. You'll need to take steps to *stub* and *mock* other modules that your module usually leverages in order to *isolate* each test. You'll typically also need to *spy* on actions that the module takes to verify that they occur.
```
stubailo: I think this should be a bulleted list with the name of the testing type in bold at the beginning. From here it's not clear that this is an emumeration of types of testing.

content/testing.md
```diff
@@ -0,0 +1,256 @@
...
+
+<h2 id="test-modes">Test modes in Meteor</h2>
+
+- Various ways to run your application that allows different testing modalities
+
+- `meteor test` simply loads your test files and allows you to import modules you want to test
+
+- `meteor test --full-app` runs your app as usual with extra test files included
+
+- `meteor test-packages` is a way of testing Atmosphere packages, which we'll discuss in more detail in the [Writing Packages Article](writing-packages.html).
```
stubailo: Is it even worth mentioning this here?

content/testing.md
```diff
@@ -0,0 +1,256 @@
...
+
+  - console-reporters that run completely on the command-line and are primary used for automated testing like [continuous integration](#ci) (as we'll see, typically PhantomJS is used to drive such tests).
+
+While developing your app, chances are you'll want to run unit tests against a web reporter; for our example we will use a reporter that renders to Mocha's default web UI. We can add the driver simply by adding the [`avital:mocha`](https://atmospherejs.com/avital/mocha) package to our app.
+
+```bash
+meteor add avital:mocha
+```
+
+This package also doesn't do anything in development or production mode, but when our app is run in [test](#test-mode) or [full-app](#full-app-test-mode) test mode, it takes over, running test code on both the client and server, and rendering results to the browser.
```
stubailo: I don't think it's worth separating these so much - we should just call it "test mode" and then "full-app" is a subset of that.

content/testing.md
```diff
@@ -0,0 +1,256 @@
...
+
+In this article, we'll use the popular [Mocha](https://mochajs.org) test runner alongside the [Chai](http://chaijs.com) assertion library to test our application. In order to write tests in Mocha, we can add the [`avital:mocha`](https://atmospherejs.com/avital/mocha) package to our app.
+
+```
+meteor add avital:mocha
+```
+
+<h3 id="testing-externally">Testing externally</h3>
```
stubailo: I feel like end-to-end tests is a better name? I don't really know what "external" means for tests

content/testing.md
```diff
@@ -0,0 +1,256 @@
...
+
+- Using factories (`dburles:factory`)
+
+- Stubbing collections w/ factories
+
+<h2 id="unit-testing">Unit testing</h2>
+
+Unit testing is the process of isolating a module of code and then testing that the internals of that module work as you expect. As [we've split our code base up into ES2015 modules](structure.html) it's natural to test those modules one at a time.
+
+By isolating a module and simply test its internal functionality, we can write tests that are *fast* and *accurate*---they can quickly tell you where a problem in your application lies. Note however that incomplete unit tests can often hide bugs because of the way they stub out dependencies. For that reason it's useful to combine unit tests with slower (and perhaps less commonly run) integration and end-to-end tests.
```
stubailo: accurate -> precise

content/testing.md
```diff
@@ -0,0 +1,256 @@
...
+class TodosCollection extends Mongo.Collection {
+  insert(doc, callback) {
+    doc.createdAt = doc.createdAt || new Date();
+    return super(doc, callback);
+  }
+}
+
+export default Todos = new TodosCollection('Todos');
+```
+[`imports/todos/Todos.js`]
```
stubailo: What's this - are we trying to link to the source code?

content/testing.md
```diff
@@ -0,0 +1,256 @@
...
+
+We should test that the `Todos` collection indeed behaves as we expect and sets that field when we insert a doc. To do that, we can write a file `imports/todos/todos.tests.js`, and define a Mocha test in it:
+
+```js
+import { mocha } from "avital:mocha";
+import { chai } from "practicalmeteor:chai";
```
stubailo: This code sample will certainly need to be updated.

stubailo:
@tmeasday looks reasonable to me, I think it's really hard for me to imagine what the final version will look like though. There is still a lot of work to do on this article and some decisions to make.

tmeasday:
@stubailo would you mind taking another look over this? Apart from the CI section, which is lacking technical detail as I'm waiting on a couple of things, it's pretty much done as a first draft.

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+
+In this article, we'll use the popular [Mocha](https://mochajs.org) test runner alongside the [Chai](http://chaijs.com) assertion library to test our application. In order to write tests in Mocha, we can add the [`avital:mocha`](https://atmospherejs.com/avital/mocha) package to our app.
+
+```bash
+meteor add avital:mocha
+```
+
+This package also doesn't do anything in development or production mode (in fact it declares itself `testOnly` so it is not even included in those modes), but when our app is run in [test mode](#test-modes), it takes over, executing test code on both the client and server, and rendering results to the browser.
+
+Test files themselves (files named `*.[app]-test[s].*` or `*.[app]-spec[s].*`) can register themselves to be run by the test driver in the usual way for that testing library. For Mocha, that's by using `describe` and `it`:
```
stubailo: These regex patterns might be confusing for people who aren't familiar for that syntax. Perhaps, give an example file name.

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+});
+```
+
+This technique will only work on the server. If you need to reset the database from a client test, you can use a method to do so:
+
+```js
+import { resetDatabase } from 'meteor/xolvio:cleaner';
+
+Meteor.methods({
+  'test.resetDatabase': () => resetDatabase();
```
stubailo: We better make sure people don't put this in their production code! Maybe add a comment to that effect.

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+const todo = Factory.create('todo');
+
+// If we have a list already, we can pass in the id and avoid creating another:
+const list = Factory.create('list');
+const todoInList = Factory.create('todo', { listId: list._id });
+```
+
+<h3 id="mocking-the-database">Mocking the database</h3>
+
+As `Factory.create` directly inserts documents into the collection that's passed into the `Factory.define` function, it can be a problem to use it on the client. However there's a neat isolation trick that you can do to replace the server-backed `Todos` [client collection](collections.html#client-collections) with a mocked out [local collection](#collections.html#local-collections), that's encoded in the `stub-collections` package (currently a local package in the Todos example application).
```
stubailo: We should come up with how to call out these "fake" testing packages. Maybe explicitly say - they work, but we hope someone will write a better one, or something?

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+import { Meteor } from 'meteor/meteor';
+import { Factory } from 'meteor/factory';
+import { Random } from 'meteor/random';
+import { chai } from 'meteor/practicalmeteor:chai';
+import { StubCollections } from 'meteor/stub-collections';
+import { Template } from 'meteor/templating';
+import { _ } from 'meteor/underscore';
+import { $ } from 'meteor/jquery';
+import { FlowRouter } from 'meteor/kadira:flow-router';
+import { sinon } from 'meteor/practicalmeteor:sinon';
```
stubailo: Is it important to use a Meteor-specific sinon? Or can we use it from NPM or something?

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+  const poll = Meteor.setInterval(() => {
+    if (DDP._allSubscriptionsReady()) {
+      clearInterval(poll);
+      done();
+    }
+  }, 200);
+};
+
+// Utility to allow throwing inside callback
+const catchAsync = (done, cb) => () => {
```
stubailo: I have no idea what this is doing.

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+  try {
+    cb();
+  } catch (e) {
+    done(e);
+  }
+};
+
+if (Meteor.isClient) {
+  describe('data available when routed', () => {
```
stubailo: I think this blob could use _a lot_ more comments - it's very dense.

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+
+Of note here:
+
+ - Before running, each test sets up the data it needs using the `generateData` helper (see [the section on creating integration test data](#creating-integration-test-data) for more detail) then goes to the homepage.
+
+ - Although Flow Router doesn't take a done callback, we can use `Tracker.afterFlush` to wait for all it's reactive consequences to occur.
+
+ - Here we wrote a little utility (which could be abstracted into a general package) to wait for all the subscriptions which are created by the route change (the `todos.inList` subscription in this case) to become ready before checking their data.
+
+ - The `catchAsync` utility is useful because otherwise we don't see the results of failures inside callbacks. XXX: perhaps this test should be rewritten with promises to avoid this?
```
stubailo: Yeah this is weird.

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+
+
+<h3 id="running-full-app-tests">Running full-app tests</h3>
+
+To run the [full-app tests](#test-modes) in our application, we run:
```
stubailo: Question - can I run 3 copies of my app? app, test, and full-app test?

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+
+```
+meteor test --full-app --driver-package avital:mocha
+```
+
+When we connect to the test instance in a browser, we want to render a testing UI rather than our app UI, so the `mocha-web-reporter` package will hide any UI of our application and overlay it with its own. However the app continues to behave as normal, so we are able to route around and check the correct data is loaded.
+
+<h3 id="creating-integration-test-data">Creating data</h3>
+
+To create test data in full-app test mode, it usually makes sense to create some special test methods which we can call from the client side. Similar to the way we cleared the database using a method in the `beforeEach` in the [test data](#test-data) section above, we can call that method before running our tests.
```
stubailo: I'm not sure why this is different from creating data in regular test mode. I think each mode we cover should elaborate on why it can do things the previous one can't, and what should be done differently and why.

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+  },
+});
+
+let generateData;
+if (Meteor.isClient) {
+  // Create a second connection to the server to use to call test data methods
+  // We do this so there's no contention w/ the currently tested user's connection
+  const testConnection = Meteor.connect(Meteor.absoluteUrl());
+
+  generateData = () => Promise.denodeify((cb) => {
```
stubailo: This is very complicated looking. I think we should dial down the arrow functions, and at least add some more parens/brackets.

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+    testConnection.call('generateFixtures', cb);
+  })();
+}
+
+export { generateData };
```
stubailo: Is this a trick to export something only on the client?

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+
+```json
+{
+  "scripts": {
+    ...
+    "chimp-watch": "chimp --ddp=http://localhost:3000 --watch --mocha --path=tests",
+    "chimp-test": "chimp --mocha --path=tests"
+  },
+```
+[`package.json`]
```
stubailo: Is this some syntax for indicating the file name? How does it work?

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+    assert.equal(countLists(), initialCount + 1);
+  });
+});
+```
+
+<h3 id="running-acceptance-tests">Running acceptance tests</h3>
+
+To run the acceptance test, we first start our meteor server with a special test driver, `tmeasday:acceptance-test-driver`. (You'll need to `meteor add` it to your app). 
+
+This test driver literally does nothing, but by running our app in full app test mode, we make all of our [test data creating methods](#creating-integration-test-data) available:
```
stubailo: This was confusing for a while - why add a package that literally does nothing! But I guess what you mean is, it's impossible to run `meteor test` without a driver package, so we have to supply a blank one?

stubailo:
This is really dope! I still would love to eventually see how to test a method, how to test a publication, etc. but we can add that another time.

content/testing.md
tmeasday: I have something about that lower down.

content/testing.md
tmeasday: We have it already as it's implied by `pm:mocha`. Otherwise I'd have used it from NPM

content/testing.md
tmeasday: Just making sure if you throw from inside async code it actually gets registered by the test driver (it's weird that you have to do this). See my promise note below

content/testing.md
tmeasday: Yes (although it tends to chew the CPU a bit). Not sure whether to mention this

content/testing.md
tmeasday: Yeah this is the downside of using promises -- Meteor code isn't really setup for them -- as we know from `testAsyncMulti` there isn't really a great way to write async code in Meteor..

content/testing.md
stubailo: Right, my point is that a reader wouldn't be able to figure this out, since it's very technical JS plumbing code.

content/testing.md
tmeasday: Yeah, this was a relatively late idea--I'm not sure if I should include it, but it's definitely handle to be able to load test scaffolding when doing acceptance tests.

content/testing.md
tmeasday: Yeah, I mean it just exports `undefined` on the server.

content/testing.md
tmeasday: Well actually in this case the code is more complex than it needs to be though :)

tmeasday:
```
  generateData = Promise.denodeify((cb) => {
    testConnection.call('generateFixtures', cb);
  });
```

stubailo: what's denodeify? It makes a callback that accepts `err, res`?

content/testing.md
tmeasday: :+1:

content/testing.md
tmeasday: :

content/testing.md
tmeasday: I'm going to rewrite this with promises after all.

content/testing.md
tmeasday: denodify is a basic node-style-async -> promise utility

content/testing.md
tmeasday: Nope, just trying to indicate the filename, I'll move it to the text

content/testing.md
tmeasday: Let me know what you think now

content/testing.md
tmeasday: Got rid of it

content/testing.md
tmeasday: I guess it's not, but it doesn't necessarily make sense to generate data on the server when running a client test in non-full-app mode, as you won't have publications etc

tmeasday:
Addressed some of your comments, made a few notes

content/testing.md
stubailo: I think it's fine, but I would first list how to run acceptance tests without it, and then add it as a tip for later?

content/testing.md
stubailo: Right - I mean we should call that out explicitly. Since we have one of these sections in each testing mode, it's a good place to call out why it's different from the previous once.

stubailo:
Going to actually run the guide app and see what it looks like.

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+
+ - If you want to write a test that can be run against any running version of your app and verifies at the browser level that the right things happen when you push the right buttons, then you are writing an **acceptance test** (sometimes called "end to end test". Such tests typically try to hook into the application as little as possible, beyond perhaps setting up the right data to run a test against.
+
+ - Finally you may wish to test that your application works under typical load or see how much load it can handle before it falls over. This is called a **load test** or **stress test**. Such tests can be challenging to set up and typically aren't run often but are very important for confidence before a big production launch.
+
+<h3 id="challenges-with-meteor">Challenges of testing in Meteor</h3>
+
+In some ways, testing a Meteor is no different to testing any other client-server, JavaScript heavy application. However, especially compared to a more backend focussed, traditional framework, the client-server divide and the reactive nature of code can add extra challenges to testing.
```
stubailo: I think this would be easier to consume if it were a more structured bulleted list. These are great things to mention.

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+
+XXX: say more about this?
+
+<h2 id="test-modes">Test modes in Meteor</h2>
+
+The primary way to test your application in Meteor is the `meteor test` command. 
+
+This loads your application in a special "test mode". What this does is:
+
+ 1. *Doesn't* eagerly load *any* of our application code as Meteor normally would.
```
stubailo: For some reason this list doesn't seem as clear as it could be. Maybe I'll try editing it a bit. But it's critical that people understand exactly what is going on here.

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+Additionally, Meteor offers a "full application" test mode. You can run this with `meteor test --full-app`.
+
+This is similar to test mode, with key differences:
+
+ 1. It loads test files matching `*.app-test[s].*` and `*.app-spec[s].*`.
+ 2. It **does** eagerly load our application code as Meteor normally would.
+
+This means that the entirety of your application (including for instance the web server and client side router) is loaded and will run as normal. This enables you to write much more [complex integration tests](#full-app-integration-test) and also load additional files for [acceptance tests](#acceptance-testing).
+
+Note that there is another test command in the Meteor tool; `meteor test-packages` is a way of testing Atmosphere packages, which we'll discuss in more detail in the [Writing Packages Article](writing-packages.html#testing).
```
stubailo: Remove this reference

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+
+There are two main kinds of test driver packages:
+
+  - web-reporters which are Meteor applications and display a special test reporting web UI that you can view the test results in.
+
+  <img src="images/mocha-test-results.png">
+
+  - console-reporters that run completely on the command-line and are primary used for automated testing like [continuous integration](#ci) (as we'll see, typically PhantomJS is used to drive such tests).
+
+In this article, we'll use the popular [Mocha](https://mochajs.org) test runner alongside the [Chai](http://chaijs.com) assertion library to test our application. In order to write tests in Mocha, we can add the [`avital:mocha`](https://atmospherejs.com/avital/mocha) package to our app.
```
stubailo: I think this warrants its own heading - maybe "Using Mocha" or something?

content/testing.md
```diff
@@ -0,0 +1,683 @@
...
+
+This technique will only work on the server. If you need to reset the database from a client test, you can use a method to do so:
+
+```js
+import { resetDatabase } from 'meteor/xolvio:cleaner';
+
+// NOTE: Before writing a method like this you'll want to double check this file is only going 
+// to be loaded in test mode!!
+Meteor.methods({
+  'test.resetDatabase': () => resetDatabase();
```
stubailo: Methods shouldn't be written using arrow functions because they often depend on `this`. It's not a big deal here, but we might not want people to get into the habit of using arrow functions and methods together.

stubailo:
OK I made a few changes. We should do another round of fine-grained editing before merging (once CI is done?) there are some more typos and stuff I'd like to fix.

tmeasday:
Let's call this done now @stubailo

stubailo:
There were still a few XXXs and stuff when I just read it. might be worth giving it just one more complete read before merging - up to you!

tmeasday:
The only XXX I'd be concerned about there was the tracker one. I think I'll leave it for now.

I'm merging, but I'm going to proof all the new articles today.

PR #253, opened 2016-02-29
Lists.private returns private fields

jholl:
Going by the example, did you intent to return an undisclosed `Lists.privateFields`?


stubailo:
No no - lists.private is meant to return the private lists. The point of private fields is that they wouldn't be sent to the client at all.

jholl:
okie dokie; I wasn't sure the context of the two uses of public/private in ea block.  Since one might choose to send different fields on different publications, I that you may have been intending to demonstrate that variance. closing.

PR #256, opened 2016-03-02
First pass at React guide

tmeasday:
Let me know what you think, especially about covering the most important ground from http://react-in-meteor.readthedocs.org/en/latest/

### DIFF
--- a/content/react.md
+++ b/content/react.md
@@ -57,8 +57,8 @@ You need to include a `<div id="app"/>` in your body's HTML somewhere of course.
 
 If you are not planning on [using both React and Blaze](#using-with-blaze) together, you can remove Blaze from your project (it's automatically added to all new projects), by running:
 
-```
-meteor remote blaze-html-templates
+```sh
+meteor remove blaze-html-templates
 meteor add static-html
 ```
 
--- a/content/react.md
+++ b/content/react.md
@@ -1,0 +1,271 @@
+---
+title: React
+order: 8
+description: How to use React, the frontend rendering library with Meteor.
+---
+
+After reading this guide, you'll know:
+
+1. What React is, and why you would consider using it with Meteor.
+2. How to install React in your Meteor application, and how to use it correctly.
+3. How to integrate React with Livedata, Meteor's realtime data layer.
+4. How to route in a React/Meteor application.
+
+<h2 id="introduction">Introduction</h2>
+
+[React](https://facebook.github.io/react/) is a JavaScript library for building reactive user interfaces developed and distributed by the Facebook team. React is one of the three rendering libraries supported by Meteor; it is an alterative to [Blaze](blaze.html) and [Angular](angular.html).
+
+React is has a vibrant and growing ecosystem and is used widely in production in a variety of combinations with different frameworks.
+
+To learn more about using React in general and coming up to speed with the library, you should check out the [React documentation](https://facebook.github.io/react/docs/getting-started.html), especially the [thinking in React](https://facebook.github.io/react/docs/thinking-in-react.html) post, which explains the React philosophy well.
+
+To get started with React in Meteor, you can follow along the [React tutorial](https://www.meteor.com/tutorials/react/creating-an-app). To see an example of a more complete Meteor application built with React, check out the [`react` branch](https://github.com/meteor/todos/tree/react) of the Todos exmaple application. Where applicable, code examples in this article will reference that app.
+
+<h3 id="using-with-meteor">Installing and using React</h3>
+
+To install React in Meteor 1.3 you should simply add it as an NPM dependency:
+
+```
+npm install --save react react-dom
+```
+
+This will install `react` into your project and allow you to access it within your files with `import React from 'react'`. Most React code is written in [JSX](https://facebook.github.io/react/docs/jsx-in-depth.html), which you can use by [default in Meteor](http://guide.meteor.com/build-tool.html#react-jsx) if you include the `ecmascript` package (which is installed in all Meteor apps by default).
+
+```js
+import React from 'react';
+
+export default class HelloWorld extends React.Component {
+  render() {
+    return <h1>Hello World</h1>;
+  }
+}
+```
+
+You can render a component heirarchy to the DOM using the `react-dom` package:
+
+```js
+import { Meteor } from 'meteor/meteor';
+import { render } from 'react-dom';
+import HelloWorld from './HelloWorld.jsx';
+
+Meteor.startup(() => {
+  render(HelloWorld, document.getElementById('app'));
+});
+```
+
+XXX: should we be getting rid of `blaze-html-templates` and replacing with `static-html`? The tutorial doesn't sem to do it..
+
+<h3 id="using-third-party-npm-packages">Using 3rd party packages</h3>
+
+If you'd like to use a third party React component that has been published on NPM (such as the ones you find on the [React Components site](http://react-components.com)), you can simple `npm install --save` them and `import` from within your app.
+
+If you are looking to write a Atmosphere package that wraps such a component, you need to take some [further steps](#atmosphere-packages).
+
+<h3 id="using-with-blaze">Using Blaze with React</h3>
+
+If you'd like to use React within a larger app built with [Blaze](#blaze.html), you can use the [`react-template-helper`](https://atmospherejs.com/meteor/react-template-helper) component which renders a react component inside a Blaze template. First run `meteor add react-template-helper`, then user the `React` helper in your template:
+
+```html
+<template name="userDisplay">
+  <div>Hello, {{username}}</div>
+  <div>{{> React component=UserAvatar userId=_id}}</div>
+</template>
+```
+
+You will need to pass in the component class with a helper:
+
+```js
+import { Template } from 'meteor/templating';
+
+import 'userDisplay.html';
+import UserAvatar from './UserAvatar.jsx';
+
+Template.userDisplay.helpers({
+  UserAvatar() {
+    return UserAvatar;
+  }
+})
+```
+
+The `component` argument is the React component to include, which should be passed in with a helper.
+
+Every other argument is passed as a prop to the component when it is rendered.
+
+Note that there a few caveats:
+
+ - React components must be the only thing in the wrapper element, due to a limitation of React (see facebook/react [#1970](https://github.com/facebook/react/issues/1970), [#2484](https://github.com/facebook/react/issues/2484)), a React component must be rendered as the only child of its parent node, meaning it cannot have any siblings.
+
+ - This means a component also can't be the only thing in a template, because it's impossible to tell where the template will be used.
+
+<h4 id="passing-callbacks-from-blaze">Passing callbacks to a React component</h4>
+
+To pass a callback to a React component that you are including with this helper, simply make a [template helper that returns a function](http://guide.meteor.com/blaze.html#pass-callbacks), and pass it in as a prop, like so:
+
+```js
+Template.userDisplay.helpers({
+  onClick() {
+    var self = Template.instance();
+
+    // Return a function from this helper, where the template instance is in
+    // a closure
+    return function () {
+      self.hasBeenClicked.set(true)
+    }
+  }
+});
+```
+
+To use it in Blaze:
+
+```js
+<template name="userDisplay">
+  <div>{{> React component=UserAvatar userId=_id onClick=onClick}}</div>
+</template>
+```
+
+<h2 id="livedata">Using data</h2>
+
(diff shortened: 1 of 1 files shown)


content/react.md
```diff
@@ -0,0 +1,271 @@
+---
+title: React
+order: 8
+description: How to use React, the frontend rendering library with Meteor.
```
stubailo: This is awk - it makes it sound like React is Meteor's frontend rendering lib

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+---
+
+After reading this guide, you'll know:
+
+1. What React is, and why you would consider using it with Meteor.
```
stubailo: Maybe backwards? Like "Why you should consider using Meteor as a backend for a React app"? Not sure what the target audience is TBH

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+2. How to install React in your Meteor application, and how to use it correctly.
+3. How to integrate React with Livedata, Meteor's realtime data layer.
```
stubailo: Do we actually call it livedata anywhere?

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+```
+npm install --save react react-dom
+```
+
+This will install `react` into your project and allow you to access it within your files with `import React from 'react'`. Most React code is written in [JSX](https://facebook.github.io/react/docs/jsx-in-depth.html), which you can use by [default in Meteor](http://guide.meteor.com/build-tool.html#react-jsx) if you include the `ecmascript` package (which is installed in all Meteor apps by default).
+
+```js
+import React from 'react';
+
+export default class HelloWorld extends React.Component {
```
stubailo: Not a fan of `default` but YMMV

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+  render() {
+    return <h1>Hello World</h1>;
```
stubailo: Canonical style is to wrap JSX in parens

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+
+You can render a component heirarchy to the DOM using the `react-dom` package:
+
+```js
+import { Meteor } from 'meteor/meteor';
+import { render } from 'react-dom';
+import HelloWorld from './HelloWorld.jsx';
+
+Meteor.startup(() => {
+  render(HelloWorld, document.getElementById('app'));
```
stubailo: I thought this needed to be `render(<HelloWorld />, ...)`

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+});
+```
+
+XXX: should we be getting rid of `blaze-html-templates` and replacing with `static-html`? The tutorial doesn't sem to do it..
```
stubailo: The tutorial doesn't do it because static-html didn't exist yet I think? And we end up using Blaze for accounts-ui anyway. So I'm not sure.

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+
+<h3 id="using-third-party-npm-packages">Using 3rd party packages</h3>
+
+If you'd like to use a third party React component that has been published on NPM (such as the ones you find on the [React Components site](http://react-components.com)), you can simple `npm install --save` them and `import` from within your app.
```
stubailo: Probably worth having an example here, like react griddle or something

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+
+If you are looking to write a Atmosphere package that wraps such a component, you need to take some [further steps](#atmosphere-packages).
+
+<h3 id="using-with-blaze">Using Blaze with React</h3>
```
stubailo: Probably worth mentioning that this can be a migration strategy if you have a Blaze app that you want to move to React?

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+To pass a callback to a React component that you are including with this helper, simply make a [template helper that returns a function](http://guide.meteor.com/blaze.html#pass-callbacks), and pass it in as a prop, like so:
+
+```js
+Template.userDisplay.helpers({
+  onClick() {
+    var self = Template.instance();
+
+    // Return a function from this helper, where the template instance is in
+    // a closure
+    return function () {
```
stubailo: Maybe use arrow functions and `this`, instead of `self`?

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+
+To use it in Blaze:
+
+```js
+<template name="userDisplay">
+  <div>{{> React component=UserAvatar userId=_id onClick=onClick}}</div>
+</template>
+```
+
+<h2 id="livedata">Using data</h2>
```
stubailo: This title is a bit vague, I think it should be improved to maybe "data loading" or "integration with Meteor data system" or something

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+
+(Note that "container components" are analogous to the "smart components" and "presentational components" to the "reusable components" in the pattern we document in the [UI/UX article](http://guide.meteor.com/ui-ux.html#components), if you'd like to read more about how this philosophy marries with Meteor).
+
+For example, in the Todos example app, we have a `ListPage` component, which renders the metadata about a list alongside the todos that are within it. In order to do so, it needs to [subscribe](data-loading.html#subscriptions) to the `todos.inList` publication, check that subscription's readiness, then fetch the list of todos from the `Todos` collection.
+
+It also needs to be responsive to reactive changes in the state of those actions (for instance if a todo changes due to the action of another user). All this data loading complexity is a typical use-case for a container-presentational component split, and the `createContainer()` function makes it simple to do this.
+
+We simply define the `ListPage` component to be a presentational component that expects it's data to be passed in as a property:
```
stubailo: I thought in our convention for Blaze, `ListPage` was actually a smart component?

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+
+ListPage.propTypes = {
+  list: React.PropTypes.object,
+  todos: React.PropTypes.array,
+  loading: React.PropTypes.bool,
+  listExists: React.PropTypes.bool,
+};
+```
+
+Then we create a `ListContainer` component which wraps it and provides a data source:
```
stubailo: Maybe `ListPageContainer`, then?

Also, I think `export default` obfuscates the name of the container, I'd much prefer `export const ListPageContainer =` myself...

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+  return {
+    loading,
+    list,
+    listExists,
+    todos: listExists ? list.todos().fetch() : [],
+  };
+}, ListPage);
+```
+
+Note that the `MeteorDataContainer` container created by `createContainer()` will be fully reactive to any changes to [reactive data sources](https://atmospherejs.com/meteor/tracker) call from inside the function provided to it.
```
stubailo: I feel like the symbol `MeteorDataContainer` came out of nowhere here. What is that, an internal class used by `createContainer`?

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+
+<h3 id="preventing-rerenders">Preventing re-renders</h3>
+
+Sometimes the Tracker system can lead to unnecessary re-computations, which when combined with `react-meteor-data` leads to unnecessary re-renders. Although React is in general quite efficient in such cases, if you need to control re-rendering, the above pattern allows you to easily use React's [`shouldComponentUpdate`](https://facebook.github.io/react/docs/component-specs.html#updating-shouldcomponentupdate) on the presentational component wrapped and avoid re-renders.
```
stubailo: > unnecessary re-computations

This frames it rather negatively - what defines an unnecessary computation? I guess it's "Sometimes your data triggers a re-computation, but you know it won't affect your UI component" or something?

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+
+<h2 id="routing">Routing</h2>
```
stubailo: Worth linking to the routing guide up front to say "here are some principles about routing"

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+import { Meteor } from 'meteor/meteor';
+import { render } from 'react-dom';
+import { renderRoutes } from '../imports/startup/client/routes.jsx';
+
+Meteor.startup(() => {
+  render(renderRoutes(), document.getElementById('app'));
+});
+```
+
+In general routing with React Router in Meteor follows generally the [same principles](routing.html) as using Flow Router, however you'll need to follow some of the idioms outlined in the [documentation](https://github.com/reactjs/react-router/blob/latest/docs/Introduction.md).
```
stubailo: Would be nice to have a summary of some of the different idioms for people who are trying to decide between Flow Router and React Router.

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+
+<h2 id="meteor-and-react">Meteor and React</h2>
+
+<h3 id="atmosphere-packages">Using React in Atmosphere Packages</h3>
+
+If you are writing an Atmosphere package and want to depend on React or an NPM package that itself depends on React, you can't use `Npm.depends()` and `Npm.require()`, as this will result in *2* copies of React being installed into the application (and besides `Npm.require()` only works on the server).
```
stubailo: I thought Npm.require actually works on the client now?

content/react.md
```diff
@@ -0,0 +1,271 @@
...
+
+Instead, you need to ensure that users of your package have installed the correct packages at the application level. This will ensure a *single* copy of React is shipped to the client and all versions line up.
```
stubailo: "ensure" -> "check"

content/react.md
tmeasday: IDK, it's in the middle of the meteor guide, I'm not sure it reads well to pitch meteor

content/react.md
tmeasday: :+1:

content/react.md
tmeasday: :+1:

content/react.md
tmeasday: I think you're right.

content/react.md
tmeasday: :+1:

content/react.md
tmeasday: Got it. I've added a sentence here.

content/react.md
tmeasday: :+1:

content/react.md
tmeasday: :+1:

content/react.md
tmeasday: Can't use this but updated to ES6

content/react.md
tmeasday: Things are implemented slightly differently in the react branch. I'm not sure if it's worth refactoring it to exactly mirror the Blaze structure (like ideally I'd do it, but...)

content/react.md
tmeasday: I wanted to call it something but it is redundant anyway. :

tmeasday: :+1:

content/react.md
tmeasday: :+1:

content/react.md
tmeasday: Nope

tmeasday: `require()` does though.

content/react.md
stubailo: IDK - I feel like the Guide should be correct and it's fine if the example app is slightly wrong?

tmeasday: Yeah, good idea. I've updated it to match what we do in Blaze

tmeasday:
What does that ambiguous :+1:  mean? ;)

content/react.md
stubailo: :+1:

content/react.md
stubailo: :+1:

stubailo:
Oh weird, that was actually on the commit where you said "improvements as per sashko" to indicate my approval of the commit.

tmeasday:
Oh, gotcha. Shall we merge this now?

stubailo:
@tmeasday we need to add a code sample of how the container could be used. right now there is the `params: { id }` argument and I don't even know what I need to pass in there.

tmeasday:
Oh right, yeah that comes from the React Router. I'm not quite sure how much detail to go into about this, given half of our users won't understand RR.

stubailo:
I mean I still think it's crucial to know how the props of a container component get passed into that function. So we can use a toy example if you want, but that seems like a pretty important part of the docs for the react container. Unless we plan on having detailed docs somewhere else.

tmeasday:
Well we definitely plan on having detailed docs elsewhere. (I guess a React DOCS.md that goes into the main docs.meteor.com?). 

But I think you are right. Maybe we should just say something like:

---

Although this `ListsShowPage` is intended to be instanciated by the React Router (which passes in the props automatically), if we did ever want to create one manually, we would need to pass in the props to the container component:

```js
<ListsShowPage params={{id: '7'}}/>
```

stubailo:
Yeah. It's also weird that the arguments use a relatively obscure ES2015 feature (nested destructuring)

tmeasday:
Let's change that for clarity then.

tmeasday:
better?

stubailo:
Looks awesome man

PR #259, opened 2016-03-06
error in js code, kws is not a variable

rolandkofler:
replaced a nonexisting variable kws with the parameter options, which seems the logic argument here.


tmeasday:
Thanks!

PR #262, opened 2016-03-07
First pass at notes for React migration guide

tmeasday:


content/1.3-migration.md
```diff
@@ -89,6 +91,40 @@ Along side some of the breaking mobile changes [listed above](#breaking-mobile),
 
 - The plugin now allows for local file access on both iOS and Android. You can construct file system URLs manually (`http://localhost:<port>/local-filesystem/<path>`) or use `WebAppLocalServer.localFileSystemUrl()` to convert a `file://` URL. On iOS, core plugins like `cordova-plugin-file` and `cordova-plugin-camera` (using `Camera.DestinationType.FILE_URI`) will now automatically return these URLs instead of `file://` URLs.
 
+<h2 id="react">Recommended changes: React</h2>
+
+As mentioned in the [breaking changes](#breaking-changes), the React Atmosphere package now "peer" depends on the React npm packages (see the [instructions from package authors](#packages) above for more details on this). If you are using the `react` Atmosphere package (and thus the `react-runtime` package), this means you are required to `npm install` the following set of NPM packages into your app (read the [Using Packages](using-packages.html) article for more details about how to manage your NPM dependencies):
```
stubailo: Are we officially calling this a peer dependency? That seems misleading because it means something very specific in NPM.

tmeasday: Good question. I think technically it's the same thing as a peer dependency---in both cases your code will not run if you do not install the peer dependency--- it's just that in NPM actually enforces that for you, and helps with resolving dependencies (maybe?), whereas here the package craps out with an error. 

So (from the user's perspective) it's effectively the same thing. What would you call it?

stubailo: I would just say something like "expects React to be installed in the app". I just feel like this is one of the more confusing parts of Meteor 1.3 so is worth getting the explanation right.

tmeasday: Better?

stubailo:
Do the following packages still depend on `react-runtime`?

- `react-template-helper`
- `react-meteor-data`

stubailo:
I made some changes to make the first sentence more meaningful. Also, I think we should clean up the table of contents when we are done with this article

tmeasday:
> Do the following packages still depend on react-runtime?

> react-template-helper
> react-meteor-data

They probably do, but they should not. As I mentioned [here](https://github.com/meteor/react-packages/pull/161#issuecomment-193484017) they should "peer npm" depend.

Issue #234, opened 2016-02-11
Client-side state management


abhiaiyer91:
I'd be happy to submit a draft if you guys had any thought into doing this.

stubailo:
Can I rename this issue to "Guide article on managing client-side state"? I think focusing around a specific technology before thinking about the problems is a mistake.

tmeasday:
I'm in two minds on this one. I think Redux is super useful to a lot of people and there'd be a lot of value in documenting how best to use it with Meteor.

However I'm not sure that we actively recommend people use it for client-side state do we? It seems like overkill to me for most apps.

Perhaps as you suggest @stubailo we have a client-side-state article, which starts with the stores section here (http://guide.meteor.com/data-loading.html#stores) [which always seemed out of place], and then moves on to Redux as an "advanced" solution for complex apps.

abhiaiyer91:
IM in

stubailo:
Here's a folder of outlines you can use as an example: https://github.com/meteor/guide/tree/master/outlines

Let's work on an outline, and that will help us see if an article makes sense!

Please submit a PR with an outline to that directory. Excited to see what you come up with.

skirunman:
I think this is a key add to the guide as managing local state, especially for UI, should have some opinionated best practices such as don't use sessions, use template reactive-vars or reactive-dict for more global state that needs to exist across hot reloads, etc.

See https://www.discovermeteor.com/blog/study-plan-meteor-1-3/#local-state-management for some ideas I would like to see covered.

skirunman:
Related to this are https://github.com/meteor/guide/issues/77 https://github.com/meteor/guide/issues/78 https://github.com/meteor/guide/issues/79

skirunman:
There used to be some good info in the concepts section of the [docs on reactivity](https://github.com/meteor/meteor/blob/devel/docs/client/full-api/concepts.html#L278). I'd like to suggest that this gets combined into a single article that covers managing local state and reactive data sources. Probably also split out the [data loading stores](http://guide.meteor.com/data-loading.html#stores) info in to this new section and discuss this before diving in to pub/sub.

stubailo:
Yeah, that makes sense to me. If someone puts together an outline for this article in the spirit of the ones we did for the others, I'd be happy to review it!

Example outlines: https://github.com/meteor/guide/tree/master/meta/outlines

mitar:
Copying from #77:

[...]

PR #263, opened 2016-03-08
Some additions to the Meteor/React guide

ffxsam:
Open to discussion of course! People might want to learn more about Redux, what it is, if they need it, what advantages it brings to the table. Performance testing is a biggie, because components can often go on a re-render binge and cause problems. These tools can help measure and pinpoint trouble spots.

### DIFF
--- a/outlines/react.md
+++ b/outlines/react.md
@@ -13,10 +13,20 @@
   1. The `ReactMeteorData` mixin
   2. The `ReactMeteorDataContainer` higher order component
   3. Containers vs. presentational components [call out to UX guide]
-  4. Optimizing re-renders with `shouldComponentUpdate`
+  4. Optimizing re-renders with `shouldComponentUpdate` and [pre-bindings](https://medium.com/@esamatti/react-js-pure-render-performance-anti-pattern-fb88c101332f)
 
-6. Routing with React
-  2. FlowRouter
-    1. Using mounte
-  1. `react-router`
+4. Routing with React
+  1. FlowRouter
+    1. Using mount
+  2. `react-router`
     1. Basic demo, "spirit" of our routing guide
+
+5. Using Redux to Supplement Meteor
+  1. Advantages of Redux (rich dev/debug tools, time travel, dead simple undo/redo)
+  2. ???
+  3. Profit
+
+6. Performance Testing
+  1. React Perf Tools
+    1. Installation and Setup
+    2. Measuring Wasted Re-renders


tmeasday:
Hey @ffxsam I think for now we'll leave off covering non-Meteor-specific React content (unless you are feeling excited to write it!).

I think covering how to use Redux with Meteor would be quite interesting; I'd be excited to see a bit more of a summary of what you might cover?

rhyslbw:
Also interested in this

ffxsam:
Totally crunched for time at the moment, but let's leave this open for now. I'll come back when I have more time and we can talk about it

tomRedox:
@tmeasday I've just been through the loop of learning Redux and how to integrate it with Meteor.  The bit I found conceptually difficult was deciding how much, if anything, from the local minimongo store to bring into state.  

I wanted to be able to see how changes in the UI effected the data and to time travel on that, so I ended up copying the record being edited into state, but leaving the rest of the minimongo data (the data in lists that wasn't being edited for example) out of the state.  

This is the code I came up with for that: https://github.com/tomRedox/simpleCRM/tree/master/imports/ui/redux.  If you think that's a good approach I'd be happy to write it up.  If you don't, I'd be interested to hear your thoughts on a better one and I can incorporate it into that example app.

My code was based on [Abhi Aiyer's How we Redux series of articles](https://medium.com/modern-user-interfaces/how-we-redux-part-3-domain-890964824fec#.t1lhqdp3z), so he might be another good person to talk to about writing some words on Redux integration

tmeasday:
cc @abhiaiyer91

tmeasday:
Thanks @tomRedox, I'll take a look at that soon

ffxsam:
I only took a quick glance, but I would make a few suggestions with regards to conventions/syntax.

For thunks, IMO it's best to have a pattern like this:

* Thunk: USER_REQUEST_<action name>
  - Succeed: USER_<action name>_SUCCEEDED
  - Failed: USER_<action name>_FAILED

Also, I might have missed something (likely, since I didn't look in detail), but I'm not sure why you're using thunks this way:

https://github.com/tomRedox/simpleCRM/blob/master/imports/ui/redux/customer-list-actions.jsx

.. when there's no async actions being performed?

tomRedox:
@ffxsam that's great feedback, thank you.  I was a bit lost in the thunks to be truthful, I need to revisit it.  I was trying to move the dispatching out of the container, but the code felt like overkill as I was typing it, I'd very glad to find a way to slim it down.

ffxsam:
You're welcome! Redux (IMO) although simple, takes a while to fully grasp. I would keep all Redux dispatches in your container components, and pass stuff down to children via props. It keeps your code more maintainable.

[Here's an example](https://github.com/ffxsam/ffx-meteor-react-boilerplate/blob/example/client/actions/colors.js) of regular action creators and thunks.

tomRedox:
Thanks @ffxsam that's great, I'll have a look at changing the code tomorrow or over the weekend

abhiaiyer91:
I was plannign on writing an article for Client Side Data Management. But havent found the time

tmeasday:
Yeah! I mean Redux isn't necessarily React specific, I think the ideal situation would be a Client Side Data article (thanks for reminding me Abhi!). 

I'm not sure how quickly we'll get to it so I think this'd be a great one for external folks to work together on.

tmeasday:
Perhaps we can take this discussion here: https://github.com/meteor/guide/issues/234

tomRedox:
@ffxsam I've remembered why I was doing the thunks the way I was, I've added some notes to this issue that explain my reasoning: https://github.com/tomRedox/simpleCRM/issues/19. I'd really appreciate your thoughts if you had a moment at some point.

Joe312341:
Hello, 

As per the discussion [here](https://forums.meteor.com/t/meteor-react-official-architecture-with-1-3/) adding an example for redux would be very helpful. 

It would show a concrete example of how to implement React using Redux in Meteor. Could be treated as an appendix to the guide.

In addition sample apps implementing this architecture would be useful. They would have to definitely include testing though. I think creating a collections of apps like the one from @tomRedox could serve as good foundation if enhanced a bit more. But the most important part here is that they are sanctioned by MDG itself. 

Currently we just have a few apps that aren't fully built out (for example no testing) and are also not sanctioned by meteor and I think this leaves people with a lot of questions

tmeasday:
@Joe312341 - have you checked out https://github.com/meteor/todos/tree/react ?

tmeasday:
@ffxsam what is your latest thinking here?

ffxsam:
Funny how things change! :) I actually vote no on Redux. It's definitely useful for certain things like maintaining global UI state (e.g. "is the notification pane open or closed"). But in most cases, one can get away with just relying on React component state for things.

On the other topics:
* Pre-binding: yes, absolutely. In fact, I've created a [package to address this](https://www.npmjs.com/package/meteor-react-prebind).
* React Perf Tools: nice to have. I don't know if I've ever encountered any issues that it pointed out.. maybe once or twice. I don't know if it merits inclusion in the Meteor Guide though, honestly.

lorensr:
Thanks @ffxsam – moved those two items to #492 and #493. Would be happy to look at a PR. And there are some more React article improvement ideas in #397

PR #267, opened 2016-03-11
Fix variable name editingTodo in Blaze example

tvogels:
I believe the name of this variable should be `editingTodo`, considering the rest of the code snippet.


stubailo:
Thanks!

PR #268, opened 2016-03-11
Add ESLint-plugin-Meteor

dferber90:
This PR adds [ESLint-plugin-Meteor](https://github.com/dferber90/eslint-plugin-meteor) to the section *Installing and running ESLint*.

**Changes**
- add eslint-plugin-meteor
- remove `|| true` from npm lint script to keep proper exit code

**Open questions**
This commit adds `plugin:meteor/guide`, but there is no such rule preset yet. We'll discuss which rules should be part of it and add them then. Until then, I'll make `plugin:meteor/guide` an empty preset (all rules disabled).

As discussed in meteor/todos#88.

### DIFF
--- a/content/code-style.md
+++ b/content/code-style.md
@@ -85,19 +85,25 @@ Below, you can find directions for setting up automatic linting at many differen
 To setup ESLint in your application, you can install the following NPM packages:
 
 ```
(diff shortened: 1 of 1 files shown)


tmeasday:
Thanks!

PR #270, opened 2016-03-11
createContainer should be imported using name rather than default

ciwolsey:
```
import createContainer from 'meteor/react-meteor-data';
```
should be:
```
import { createContainer } from 'meteor/react-meteor-data';
```


tmeasday:
Thanks!

PR #283, opened 2016-03-21
Fixes link.

jiku:
Hope this points correctly. Shouldn't go to anything specific to (CS use of) require?


tmeasday:
What it's trying to point to hasn't been merged to the docs yet. Actually I'll leave this PR open and we'll fix it once the docs are deployed. Thanks!

PR #300, opened 2016-03-29
Correct name for Coffeescript

SergioB:
Was unable to search for this part of documentation, so decided to edit it to correct name so that everybody will be able to search for it.


SergioB:
I just signed Meteor CLA

stubailo:
Thanks for the edit!

PR #301, opened 2016-03-29
Add Atom's Reload as an alternative to Restarting

batjko:
image

This will suffice to enable the linting, no restart necessary.


stubailo:
Sounds great! Thanks.

PR #316, opened 2016-03-31
Add info on npm install requirements for builds

tjcampbell:
`meteor build` requires that npm dependencies are installed manually first. This isn't in the guide yet. Also added --production flag to avoid installing devDependencies.

### DIFF
--- a/content/deployment.md
+++ b/content/deployment.md
@@ -109,12 +109,13 @@ You can obtain a server running Ubuntu or Debian from many generic hosting provi
 
 <h3 id="custom-deployment">Custom deployment</h3>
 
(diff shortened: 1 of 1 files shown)


tmeasday:
Thanks @tjcampbell 

1. Can you please sign the CLA? Thanks!

2. I think the final `--production` is unnecessary as the `package.json` in there is a minimal one generated by the build tool.

tjcampbell:
@tmeasday I signed the CLA right after creating the PR but I see you guys have merged anyway. Up to you with the second production flag, I agree adding it just for 'chalk' and 'eachline' is probably not worth it.

PR #321, opened 2016-03-31
Run npm install as Meteor now ships with package.json

skirunman:
I suspect MDG will likely add other NPM dependencies in the future so best to run this command before running meteor the first time. In addition, I removed `$` on install command for better consistency with the rest of the guide.

### DIFF
--- a/content/index.md
+++ b/content/index.md
@@ -25,7 +25,7 @@ On Windows?  [Download the official Meteor installer here](https://install.meteo
 On OS X or Linux?  Install the latest official Meteor release from your terminal:
 
 ```bash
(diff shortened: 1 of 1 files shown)


content/index.md
```diff
@@ -42,6 +42,7 @@ Run it locally:
 
 ```bash
 cd myapp
+npm install
```
stubailo: I think this should be `meteor npm install` if it's in the getting started section, since some people might not have NPM installed globally.

skirunman: I agree and will make the change.

stubailo:
Thanks!

PR #322, opened 2016-04-01
nectar ninja installation

danahmadi:


danahmadi:
(Nectar ninja is the notification system we're using to announce the livestream as well as other special alerts)

stubailo:
MERGED

neat!

PR #327, opened 2016-04-01
Add comment about using meteor nom install

skirunman:
This has come up a couple of times on the forum today, so maybe this will make it a bit more clear on what is `meteor npm` versus just `npm`. This same comment is also in using-packages.md.

https://forums.meteor.com/t/meteor-npm-install-vs-npm-install/20495


stubailo:
Makes sense to me!

PR #328, opened 2016-04-01
How to export a module variable in CoffeeScript

fknipp:
Maybe there might be better places in the guide to put this information, but it should be written somewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ Render multiple items in your template with `{% raw %}{{#each}}{% endraw %}`
 
 ## Changelog
 
+- 2016/04/01: Added CoffeeScript exports syntax. [PR #328](https://github.com/meteor/guide/pull/328)
 - 2016/04/01: Changed Mocha test code snippets to use function expressions instead of arrow functions, after the discussion on [Issue #318](https://github.com/meteor/guide/issues/318). [PR #323](https://github.com/meteor/guide/pull/323)
 - 2016/04/01: Added `gadicc:blaze-react-component` in a new "Blaze in React" section of the React article. [PR #325](https://github.com/meteor/guide/pull/325)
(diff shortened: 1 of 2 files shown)


stubailo:
@fknipp can you add an entry to the changelog? https://github.com/meteor/guide#changelog

fknipp:
You're welcome :)

stubailo:
Thanks!

GeoffreyBooth:
I just stumbled across this. This section is just wrong. The syntax is _not_ different in CoffeeScript; CoffeeScript supports the `import` and `export` keywords, nearly identically to ES2015. What this PR is referring to is CommonJS `module.exports`, though it’s confusing CommonJS’ global `exports` object with the `export` keyword in CoffeeScript/ES2015.

This was written up here: https://docs.meteor.com/packages/modules.html#CoffeeScript. Clearly that’s hard to find, so perhaps we can replace this section with a link to that one?

abernix:
@GeoffreyBooth That sounds fine to me.  Can you do a PR or shall I?

lorensr:
Thanks Geoffrey, fixed in abedc3bd1680e5ace0cbc18e77d032b38a630aa1

GeoffreyBooth:
@lorensr Thanks for tackling this, I haven’t had time just yet. In that commit it looks like the `exports.Lists = ListsCollection 'lists'` example is still there, after the link you added. I would think that this should become `export Lists = new ListsCollection 'lists'` or similar, i.e. using the `export` keyword (and ES2015 modules) rather than `exports` (and CommonJS). Even for CoffeeScript I think we should be encouraging the `import` and `export` keywords, and ES2015 modules, over CommonJS. Backticks aren’t required anymore.

lorensr:
Oops thanks, fixed!

PR #329, opened 2016-04-01
Mention meteor-node-stubs and fix typos

skirunman:
I think it is a good idea to explain what meteor-node-stubs does as it is installed by default.


stubailo:
Thanks!

PR #330, opened 2016-04-01
Setting order for proper navigation arrows

fknipp:
I've set the order attribute on almost all pages, so that the navigation arrows work in the way how I expect them:

← Go to the previous item of the sidebar              Go to the next item of the sidebar →

The first part (until the first horizontal line in the sidebar) goes from 0 to 9, the second part from 10 to 19, etc. 

An update to README.md is not necessary, as there is no content change.

### DIFF
--- a/content/1.3-migration.md
+++ b/content/1.3-migration.md
@@ -1,6 +1,6 @@
 ---
 title: Migrating to Meteor 1.3
-order: 20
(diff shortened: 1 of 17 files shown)


stubailo:
Thanks! I forgot they weren't just being used for the table of contents.

PR #331, opened 2016-04-02
wrong path in reload bash_profile

filipenevola:
Simple change in reload bash_profile, was **~./ **instead of **~/.**

It can confuse developers without experience with bash.


stubailo:
Thanks!

PR #334, opened 2016-04-03
Added an additional line about import rules

faysal515:

If you are adding significant new content, please take a moment to include an update to the [changelog](https://github.com/meteor/guide/blob/master/README.md#changelog) in your PR.

I had a confusion that anything would be loaded in the startup folder as there was not `Meteor.startup` block. after sometimes I found out that we need to actually implicitly define in the entry point.

### DIFF
--- a/content/structure.md
+++ b/content/structure.md
@@ -117,7 +117,7 @@ For each module defined above, it makes sense to co-locate the various auxiliary
 
 <h3 id="startup-files">Startup files</h3>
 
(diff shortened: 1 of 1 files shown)


skirunman:
I'm a bit confused by your changes as no files are eagerly (automatically) loaded from `imports/`. Does it not say earlier:

> These main.js files won't do anything themselves, but they should import some startup modules which will run immediately, on client and server respectively, when the app loads.

If you want to add some clarification here I would suggest:

> ... other code that is required on the client, forming the main entry point for the rest of the client application as imported from `client/main.js`:

faysal515:
okay then. i might had problems understand the flow. i'm closing the request.

PR #335, opened 2016-04-03
Proposal: Example using Tracker.Component

timbrandin:
Just giving an example use case with Tracker.Component which also manages subscriptions much like we could before with Blaze Templates.

https://www.npmjs.com/package/tracker-component

### DIFF
--- a/content/react.md
+++ b/content/react.md
@@ -238,6 +238,73 @@ Although this `ListsShowPage` container is intended to be instantiated by the Re
 <ListsShowPage params={{id: '7'}}/>
 ```
 
+<h3 id="using-tracker-component">Using `Tracker.Component`</h3>
+
+Once you've run `npm i --save tracker-component`, you'll be able to create components with the `Tracker.Component` class, it allows you to create a [container component](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0#.by86emv9b) which provides data to your presentational components.
+
+> Note that "container components" are analogous to the "smart components" and "presentational components" to the "reusable components" in the pattern we document in the [UI/UX article](http://guide.meteor.com/ui-ux.html#components), if you'd like to read more about how this philosophy relates to Meteor.
+
+For example, in the Todos example app, we have a `ListsShow` component, which renders list metadata and the tasks in the list. In order to do so, it needs to [subscribe](data-loading.html#subscriptions) to the `todos.inList` publication, check that subscription's readiness, then fetch the list of todos from the `Todos` collection.
+
+It also needs to be responsive to reactive changes in the state of those actions (for instance if a todo changes due to the action of another user). All this data loading complexity is a typical use-case for a container-presentational component split, and `Tracker.Component` class makes it simple to do this.
+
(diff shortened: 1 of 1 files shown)


content/react.md
```diff
@@ -238,6 +238,73 @@ Although this `ListsShowPage` container is intended to be instantiated by the Re
...
+  listExists: React.PropTypes.bool,
+};
+```
+
+Then we create a `ListsShowPage` container component which wraps it and provides a data source:
+
+```js
+import { Meteor } from 'meteor/meteor';
+import { Lists } from '../../api/lists/lists.js';
+import Tracker from 'tracker-component';
```
stubailo: Is it weird that this overrides `Tracker` from `meteor/tracker`? If you have code like `Tracker.autorun` etc. seems like it wouldn't work anymore?

In general, I feel like making the package attach itself to the `Tracker` namespace doesn't make sense. Why not just `import TrackerComponent from 'tracker-component'` instead?

content/react.md
```diff
@@ -238,6 +238,73 @@ Although this `ListsShowPage` container is intended to be instantiated by the Re
 <ListsShowPage params={{id: '7'}}/>
 ```
 
+<h3 id="using-tracker-component">Using `Tracker.Component`</h3>
+
+Once you've run `npm i --save tracker-component`, you'll be able to create components with the `Tracker.Component` class, it allows you to create a [container component](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0#.by86emv9b) which provides data to your presentational components.
```
stubailo: I feel like this paragraph should start with an explanation of why one would use `tracker-component` instead of `createContainer`. Right now, there are two issues:

1. It starts with "once you've run ..." which doesn't explain why one would install the package in the first place, and
2. The code sample doesn't seem to have big benefits over `createContainer` - is it simply the lack of build-in `subscriptionsReady`?

content/react.md
timbrandin: Yeah, the idea was just to be as simple as React.Component :+1:

content/react.md
timbrandin: 1. it was just a adjustment to the createComponent text, I guess I can expand on this.
2. that and the lack of unsubscribe when the component is unmounted.

timbrandin:
Thx, for the feedback! :) I'll have a look and see what improvements I can make. No rush with this

content/react.md
stubailo: I'm pretty sure createContainer does unsubscribe... Do you have a repro? That would be a pretty huge oversight otherwise.

stubailo:
OK, let me know if I can help!

content/react.md
timbrandin: Oh, so I didn't realize that when you stopped a computation it also stopped the subscriptions done inside, that's cool. This realization also made Tracker.Component 3 lines shorter, down to 52 now. :dancer:

tmeasday:
This seems pretty neat, but I don't see enough advantage over `react-meteor-data`'s `createContainer` to be worth adding.

PR #337, opened 2016-04-03
Improving chapter about escaping

fknipp:
After reading the explanation the first time, I put the pipe character behind opening and closing braces by mistake, but it's only necessary for `{{` and `{{{`.


stubailo:
Thanks!

Issue #336, opened 2016-04-03
Test documentation doesn't mention that tests in "tests" folder are ignored


siegesmund:
The current behavior seems to be to ignore tests in any folder called "tests". I believe that this behavior was to prevent tests being included in the build in prior Meteor versions. The documentation implies that any file that that complies with the specified naming convention will be loaded in test mode, regardless of the folder it's in. This should be clarified to avoid confusion.

> This loads your application in a special "test mode". What this does is:
> 
> Doesn't eagerly load any of our application code as Meteor normally would.
> Does eagerly load any file in our application (including in imports/ folders) that look like *.test[s].*, or *.spec[s].*
> Sets the Meteor.isTest flag to be true.
> Starts up the test driver package (see below).

skirunman:
Yes, I noticed this as well. I'm fixing this in a PR I will submit today.

stubailo:
Seems like this is fixed after that PR. Agreed it's a bit counterintuitive, but it was the best way to keep backcompat (Meteor 1.2 and before also ignored the `tests/` directory)

PR #340, opened 2016-04-04
Add back in build tool default loading order rules

skirunman:

If you are adding significant new content, please take a moment to include an update to the [changelog](https://github.com/meteor/guide/blob/master/README.md#changelog) in your PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ Render multiple items in your template with `{% raw %}{{#each}}{% endraw %}`
 
 ## Changelog
 
+- 2016/04/03: Add back in build tool default loading order rules. [PR #340](https://github.com/meteor/guide/pull/340)
 - 2016/04/01: Added CoffeeScript exports syntax. [PR #328](https://github.com/meteor/guide/pull/328)
 - 2016/04/01: Changed Mocha test code snippets to use function expressions instead of arrow functions, after the discussion on [Issue #318](https://github.com/meteor/guide/issues/318). [PR #323](https://github.com/meteor/guide/pull/323)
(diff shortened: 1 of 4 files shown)


skirunman:
This should fix #336

skirunman:
This includes some cross reference changes to mobile.md. Probably need to confirm with @martijnwalraven that the `cordova-build-override/` directory still works.

stubailo:
Yeah this looks great to me. I expect the build override still works, since that would be a big backcompat change and wasn't mentioned in http://guide.meteor.com/1.3-migration.html#breaking-changes-mobile

stubailo:
Awesome work, this is a great new addition to the guide!

martijnwalraven:
The `cordova-build-override` mechanism still works, but I didn't include it in the guide because I don't think it is ever a good solution and it can lead to a lot of trouble.

The problem is that is overrides complete files, so if you use it to write your own `config.xml` for example, you lose all configuration information set by the build process and by installed plugins, which will likely break your app. 

Cordova has a `<config-file>` tag as part of `plugin.xml` that allows you to selectively change parts of configuration files, including `config.xml`, so if you absolutely must make changes, the recommendation would be to write a dummy plugin.

stubailo:
@martijnwalraven OK, sorry I merged that if it isn't recommended! Please change it back if you think it is best.

martijnwalraven:
@stubailo: I've removed the recommendation to override `config.xml` and mentioned the plugin alternative in https://github.com/meteor/guide/commit/4397c437b4f5ea9adcdad47a71f0c69068db2f74.

PR #345, opened 2016-04-05
Reorder default load order to avoid confusion

StanLindsey:
Considering `main.` files are loaded last, they should be listed last to avoid confusion. Otherwise it appears that `lib/` is loaded next, after the `.main` files.


skirunman:
I believe the reason they are listed this way is this is the precedence of how the rules are applied if there are conflicts in the rules such as a file called `main.*` that is also an HTML file, but I'm not 100% certain.

stubailo:
Yeah, the order is intentional to indicate the priority in which the rules are applied. I'd accept a PR to make that more clear, but this isn't the way to do it.

PR #352, opened 2016-04-06
Add info for migrating from local packages to modules

reohjs:
This PR adds a few tips to the 'Recommended change: use modules' section of the 'Migrating to Meteor 1.3' article on converting an app-local packages structure to a modules-based structure.

Ref: https://forums.meteor.com/t/how-to-migrating-from-all-in-packages-to-1-3-modules/20681/3

### DIFF
--- a/content/1.3-migration.md
+++ b/content/1.3-migration.md
@@ -45,20 +45,21 @@ meteor npm install --save meteor-node-stubs
 - If you are using app-local packages to control load order and write unit tests for your application, we recommend you switch to using modules:
  - Remove code related to the [Package API](http://docs.meteor.com/#/full/packagejs) from the `package.js` files and rename them to `index.js`, 
  - Move your local packages to the `imports/` directory.
- - Add the necessary import statements to each of the modules in your packages.
+ - Add the necessary `import` statements to each of the modules in your packages.
+ - Add `export` statements to each of your packages exports.
 
 ```js
-api.use('meteor');
-// Turns into
-import { Meteor } from 'meteor/meteor';
-
 api.addFiles('file.js');
-// Turns into
+// For files that are not imported elsewhere, this turns into
 import './file.js';
 
-// imports/main.js
-// Don't forget to import the package!
-import './local_package/index';
+// Remove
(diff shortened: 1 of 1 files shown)


content/1.3-migration.md
```diff
@@ -42,8 +42,26 @@ import { EJSON } from 'meteor/ejson';
...
-- If you are using app-local packages to control load order and write unit tests for your application, we recommend you switch to using modules. You can read about our recommended structure for applications and modules in the [Application Structure article](structure.html) of the Meteor Guide, and how to test them in the [Testing article](testing.html).
+- If you are using app-local packages to control load order and write unit tests for your application, we recommend you switch to using modules:
+ - Remove code related to the [Package API](http://docs.meteor.com/#/full/packagejs) from the `package.js` files and rename them to `index.js`, 
+ - Move your local packages to the `imports/` directory.
+ - Add the necessary import statements to each of the modules in your packages.
 
+```js
+api.use('meteor');
+// Turns into
+import { Meteor } from 'meteor/meteor';
```
stubailo: I don't think there is any reason to have your `index.js` import these dependencies, right?

content/1.3-migration.md
```diff
@@ -42,8 +42,26 @@ import { EJSON } from 'meteor/ejson';
...
+
+api.addFiles('file.js');
+// Turns into
+import './file.js';
+
+// imports/main.js
+// Don't forget to import the package!
+import './local_package/index';
```
stubailo: This is a bit confusing - this is saying you should have a file called `imports/main.js`, and that should import all of your local packages? That doesn't seem completely right to me. I'd rather when people want to use their module, they import:

```js
import { myModule } from '/imports/myModule'`;
```

stubailo:
I'd also suggest talking about exports in here.

content/1.3-migration.md
reohjs: Right, meant to remove this.

reohjs:
I've updated with suggestions

stubailo:
This is definitely an improvement, thanks!

PR #353, opened 2016-04-06
Fix a few broken links.

kenken17:
The first commit did not pass the check, the second one passed, but then don't know why this PR have the previous commit. Anyway, please check and see. :)


kenken17:
I think I know how it works now. Sorry for the trouble. I am closing this PR.

PR #356, opened 2016-04-07
Clarify app structure and more import examples

skirunman:
If my merge upstream commit f7cf78b is an issue let me know. Still learning idiosyncrasies of git and github and not sure if this pollutes upstream commit history.


content/structure.md
```diff
@@ -23,15 +23,41 @@ In ES2015, you can make variables available outside a file using the `export` ke
...
+You can read about the module system in detail in the [`modules` package README](https://docs.meteor.com/#/full/modules). This package is automatically included in every new Meteor app as part of the [`ecmascript` meta-package](https://docs.meteor.com/#/full/ecmascript), so most apps won't need to do anything to start using modules right away.
 
-<h3 id="importing-from-packages">Importing from packages</h3>
+<h3 id="intro-to-import-export">Introduction to using `import` and `export`</h3>
 
-In Meteor, it is also simple and straightforward to use the `import` syntax to load npm packages on the client or server, and access the package's exported symbols as you would with any other module. You can also import from Atmosphere packages, but the import path must be prefixed with `meteor/` to avoid conflict with the npm package namespace. For example, to import `HTTP` you could use `import { HTTP } from 'meteor/http'`.
+Meteor allows you to `import` not only JavaScript in your application, but also CSS and HTML:
```
stubailo: I think it's not really clear from this what importing HTML and CSS does.

content/structure.md
```diff
@@ -23,15 +23,41 @@ In ES2015, you can make variables available outside a file using the `export` ke
...
+import './loading.html';              // import HTML from relative path
+import '/imports/ui/style.css';       // import CSS from absolute path
+```
+
+Meteor also supports the standard E2015 modules `export` syntax:
+
+```js
+export const listRenderHold = LaunchScreen.hold();  // named export
+export { Todos };                                   // named export
+export default { Lists };                           // default export
```
stubailo: I think a default export wouldn't have brackets. Like so:

```js
// Exporting
export default Lists;
export default new Collection('lists');

// Importing
import AnyName from './path';
```

content/structure.md
```diff
@@ -162,15 +195,24 @@ import '../imports/startup/server/security.js';
...
+
+```js
+import { Meteor } from 'meteor/meteor';
+import { EJSON } from 'meteor/ejson';
+```
 
 <h2 id="load-order">Default file load order</h2>
 
-Even though it is recommended that you write your application to use ES2015 modules and the `imports/` directory, Meteor 1.3 continues to support eager loading of files, using these default load order rules, to provide backwards compatibility with applications written for Meteor 1.2 and earlier.
+Even though it is recommended that you write your application to use ES2015 modules and the `imports/` directory, Meteor 1.3 continues to support eager loading of files, using these default load order rules, to provide backwards compatibility with applications written for Meteor 1.2 and earlier. You may combine both eager loading and lazy loading using `import` in a single app. Any import statements are evaluated in order they are listed in a file when that file is loaded and evaluated using these rules.
```
stubailo: "in order" => "in the order"

stubailo:
Looks really good, a few comments!

stubailo:
(BTW probably what you want to do to fix this PR is `git rebase` against the upstream master branch)

content/structure.md
skirunman: How about:

Meteor allows you to `import` not only JavaScript in your application, but also CSS and HTML if you want to control their load order:

content/structure.md
skirunman: Yep.

skirunman:
@stubailo Let me know if this looks OK now. If not, I can start over...

content/structure.md
```diff
@@ -23,15 +23,42 @@ In ES2015, you can make variables available outside a file using the `export` ke
...
+<h3 id="intro-to-import-export">Introduction to using `import` and `export`</h3>
 
-In Meteor, it is also simple and straightforward to use the `import` syntax to load npm packages on the client or server, and access the package's exported symbols as you would with any other module. You can also import from Atmosphere packages, but the import path must be prefixed with `meteor/` to avoid conflict with the npm package namespace. For example, to import `HTTP` you could use `import { HTTP } from 'meteor/http'`.
+Meteor allows you to `import` not only JavaScript in your application, but also CSS and HTML to control load order:
 
-<h3 id="using-require">Using `require`</h3>
+```js
+import '../../api/lists/methods.js';  // import from relative path
+import '/imports/startup/client';     // import module with index.js from absolute path
+import './loading.html';              // import HTML from relative path
```
tmeasday: This is maybe not quite correct. I'm not sure what the best one liner is, but it's something like "import Blaze-compiled HTML"

content/structure.md
```diff
@@ -145,7 +166,20 @@ import '../../ui/accounts/accounts-templates.js';
 // Below here are the route definitions
 ```
 
-On the server, we import various modules into our `server/main.js` to set up the server environment:
+We then import both of these files in `/imports/startup/client/index.js`:
```
tmeasday: When referencing files in the app, let's not use a leading `/`, even if it is more correct.

tmeasday:
This is looking great @skirunman

skirunman:
@stubailo Anything else you want done to this PR?

tmeasday:
LGTM!

stubailo:
Love it!

content/structure.md
```diff
@@ -162,15 +196,24 @@ import '../imports/startup/server/security.js';
...
+Even though it is recommended that you write your application to use ES2015 modules and the `imports/` directory, Meteor 1.3 continues to support eager loading of files, using these default load order rules, to provide backwards compatibility with applications written for Meteor 1.2 and earlier. You may combine both eager loading and lazy loading using `import` in a single app. Any import statements are evaluated in the order they are listed in a file when that file is loaded and evaluated using these rules.
 
 There are several load order rules. They are *applied sequentially* to all applicable files in the application, in the priority given below:
 
-1. HTML template files are *always* loaded before everything else
+1. HTML template files are **always** loaded before everything else
 2. Files beginning with `main.` are loaded **last**
 3. Files inside **any** `lib/` directory are loaded next
```
batjko: It's a bit odd to say something gets "loaded *next*", when the previous point said what will be "loaded *last*".

skirunman: Yes, it can be a bit confusing, but the order listed is the priority order that the rules are evaluated, not files are loaded, as shown in the little example below this section. If you can come up with a better way to express this I'm sure MDG will accept the change.

stubailo: Yeah I think the fact that this is so confusing is one reason why we don't recommend it anymore, and instead encourage people to use modules.

PR #359, opened 2016-04-08
Minor tweak to font size on some header tags

skirunman:
Looks better this way.


stubailo:
Note that this change will make them show up in the table of contents. If you just want the text to be bigger, we can just change the CSS to make `h4` tags appear larger?

skirunman:
I think it makes sense to have in TOC as well. They were there previously.

stubailo:
Sure!

PR #362, opened 2016-04-10
Add CSS examples for import and npm import with v1.3.2

skirunman:
General cleanup and add some more examples for 1.3.2 import of CSS type files as this is now supported. I'd wait to push this until 1.3.2 is released.


tmeasday:
Thanks @skirunman!

@stubailo: I'm glad this came up, we should discuss what to do about this, as this is a new feature in 1.3.2 -- does this mean we need a new version of the guide? How should we handle this situation?

I don't really like the idea of the 1.3 guide changing in significant ways. I think people shouldn't be aiming for a moving target.

stubailo:
> I don't really like the idea of the 1.3 guide changing in significant ways. I think people shouldn't be aiming for a moving target.

Well, this is a new feature so it's not really like "oh, you did it in this way, now do it in a different way". But I can see the case for only adding this to the guide in 1.4, just to make sure that the feature works as intended before recommending people use it?

tmeasday:
I mean, we could have a 1.3.2 guide as well. I just think we need to figure out a process for these kind of changes.

skirunman:
My take is that it would be much better to only have Meteor Guide versions at 1.2, 1.3, 1.4, etc. If you are strictly following semver than the new features to support import from NPM might be better released in a 1.4 release. At the same time, I think you can argue that the API for using CSS pre-processors has not changed and this is simply a bug fix to make them work with NPM, so there is justification for shipping then in 1.3.2 release.

Another thing you could add to the intro to the Guide is that the guide reflects the state of the latest 1.x.x release so make sure you are using that release to ensure everything works exactly as stated in the Guide.

This will probably not be a big issue to just publish these specific changes for 1.3.2 is released as people need this capability. Going forward, I don't expect you'll run into many issues like this with 1.x.x releases.

skirunman:
Also, I did see the new [published draft roadmap](https://github.com/zol/meteor/blob/new-roadmap/Roadmap.md). I think Meteor should try to follow semver as much as possible so a release that supports new mongo db versions should be 1.4 and a release supporting all NPM is going to break lots of apis so should be a 2.0 release. My 2 cents.

tmeasday:
The Meteor release number is a marketing exercise so I'm not sure we can count on it following semver unfortunately.

skirunman:
Yes, I know that Meteor does not follow semver strictly, but closing the gap as much as you can helps. For instance, I definitely would expect all all NPM release of Meteor to be a 2.0 release.

skirunman:
And can we commit this now as well?

skirunman:
People are asking questions about the node_modules import documentation.

https://github.com/meteor/meteor/issues/6037#issuecomment-211452834

tmeasday:
Great! Thanks @skirunman

PR #363, opened 2016-04-10
Usage of super() outside of constructor replaced by super.methodname()

hanskohls:
This the syntax that seemed to work in Meteor 1.2, did not build in Meteor 1.3.1. 
The change is in line with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super

### DIFF
--- a/content/collections.md
+++ b/content/collections.md
@@ -216,7 +216,7 @@ class ListsCollection extends Mongo.Collection {
 
     // Call the original `insert` method, which will validate
     // against the schema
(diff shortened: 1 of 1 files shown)


tmeasday:
Thanks @hanskohls. Would you mind signing the Meteor CLA?

stubailo:
Technically we don't need the CLA for typos, so I'm going to merge this. But good to sign it eventually.

PR #364, opened 2016-04-10
Remove self-closing div from documentation

schmidsi:

If I do this, meteor crashes with the following exception:
```
Errors prevented startup:                  
   
   While processing files with templating (for target web.browser):
   client/main.html:8: Only certain elements like BR, HR, IMG, etc. (and foreign elements like SVG) are
   allowed to self-close
   ...1>    <div id="app"/>
```
Therefore, I think it is better to not show code-examples which won't work.

Or am I missing something?

### DIFF
--- a/content/react.md
+++ b/content/react.md
@@ -56,7 +56,7 @@ Meteor.startup(() => {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


stubailo:
You're right!

PR #368, opened 2016-04-12
Link to original 'faker' package instead of fork

klaussner:
The guide links to an old fork of 'faker' that is also registered on npm (as 'Faker' instead of 'faker'). The Todos app uses the up-to-date version.


stubailo:
npm is case sensitive????? but the stats aren't, apparently, because they report exactly the same numbers.

Wow, it absolutely blows my mind that these are different packages.

klaussner:
Npm dropped support for uppercase letters in package names some time ago but I guess the 'Faker' package is just too old. 😄

Does the Meteor Bot recheck the CLA status? I signed it after creating the pull request.

stubailo:
Yeah right now the bot is weird. Thanks for the PR! Really surprised to hear this could be a thing.

PR #369, opened 2016-04-12
Fix link to eslint Meteor plugin

skirunman:
I could not find a any plugin ["plugin:meteor/guide"], but there is a ["plugin:meteor/recommended"].

https://github.com/dferber90/eslint-plugin-meteor#configuration


stubailo:
Good catch!

Issue #370, opened 2016-04-12
Should "Application structure" section follow "Code style"


skirunman:
It seems to flow better and generally make more sense to have the Code style section before diving into lots of details that are very Meteor specific on structuring your app. In addition, in the Code style section there references to npm commands without any introduction to npm, this could be added to make it flow better. I'm going to add some details in the Code style section, but I'd suggest we move this section before Application structure.

PR #371, opened 2016-04-12
Minor changes for consistency

skirunman:
We are using "meteor npm" through rest of guide so why not use it here. This is related to #370.


stubailo:
Sure, thanks!

PR #372, opened 2016-04-13
provide correct ways to use Moment.js

Hongbo-Miao:
Moment.js is special.
Please check second answer [here](http://stackoverflow.com/questions/35166168/how-to-use-moment-js-library-in-angular-2-typescript-app) for details.

So instead of using
```
import moment from 'moment';
```

The correct ways should be either of these below:
```
import * as moment from 'moment';
import moment = require('moment');
```

### DIFF
--- a/content/using-packages.md
+++ b/content/using-packages.md
@@ -57,6 +57,8 @@ import moment from 'moment';
 const moment = require('moment');
 ```
 
+This works in most cases. But since moment is special, you need use either `import * as moment from 'moment';` or `import moment = require('moment');`
+
 This imports the default export from the package into the symbol `moment`.
 
 You can also import specific functions from a package using the destructuring syntax:


batjko:
The linked discussion refers to typescript issues. I don't think what you're suggesting is necessary for use in normal Javascript. 
`import moment from 'moment'` works fine for me.

Hongbo-Miao:
@batjko oh, yeah, I think you are right, actually I am using angular2-meteor.
But this document used moment.js this special example, which mislead us in angular2-meteor project.
Maybe we can somehow improve it?

batjko:
The preference for Typescript is an Angular thing. Not sure it belongs in the Meteor guide, unless of course inside the Angular sub-section?
Ultimately, I think this special use case is better addressed in the Angular-Meteor documentation.

Hongbo-Miao:
@batjko yeah, maybe we should add into angular2-meteor document.

stubailo:
I think it would be worth it to change the example to avoid confusing people though. If moment is the only package that has this weirdness maybe it's not a good first NPM example.

Hongbo-Miao:
@stubailo sorry, it is my mistake.
Actually, in Angular 2, all `import` from NPM are similar like this:
`import * as PACKAGE from 'PACKAGE';`
instead of 
`import PACKAGE from 'PACKAGE';`

stubailo:
That's not really the case - there are lots of places where `import` and `require` have subtle incompatibilities. It's kind of like arrow functions, where it's unfortunate that there are two ways to do something that are almost the same but a tiny bit different..

batjko:
Is there any consistent pattern to those subtle differences, that would be
worth pointing out when discussing imports?

stubailo:
Probably! Although I don't know if we want to be in the business of writing documentation for how the edge cases of ES2015 import transpilation works... or maybe we do? If we could put something up that could be a good guide for the wider JS community that could be worth it.

Worth opening a new issue about it.

batjko:
I think only in so far as they relate to the Meteor stack directly or to
those de-facto standard packages that the guide refers to, say this Angular
one for example.

PR #379, opened 2016-04-15
Fix two typos in README.md

intron:
Ran across a couple of typos in the Testing section of the docs.

* "isolating" should match "testing"
* "reactive consequences" belongs to "Flow Router", so we should use "its"


stubailo:
Great, thanks!

PR #380, opened 2016-04-15
Pass data from a container into a FlowRouter route

andersr:
How to pass data from a top-level container in FlowRouter to the component used for a given route.

@stubailo  - created per your [comment](https://forums.meteor.com/t/using-data-in-meteor-with-react-and-flowrouter/21484/4?u=andersr) on the forum...pls let me know if this is what you were requesting.


stubailo:
Great, thanks! This is perfect.

PR #381, opened 2016-04-16
Add sections on creating an npm package and overriding a published one

### LINKED ISSUE #378: Local version of npm packages?
People who are used to local Atmosphere packages

http://guide.meteor.com/using-packages.html#atmosphere-overriding

may wonder what the closest equivalent is for npm?

lorensr:
The previously linked npm explanation of creating an npm package is heavyweight for a Meteor dev's purposes (and doesn't cover the common case of private code). Using their method, you need to publish and update after each change. A Meteor dev is usually creating a package in support of their app, and is usually developing it inside the app, so using source control makes more sense. Also, the source control way is more familiar coming from `packages/`.

Also fixes #378

### DIFF
--- a/content/using-packages.md
+++ b/content/using-packages.md
@@ -251,7 +251,9 @@ Now `left-pad` is included in your `package.json`, and the code has been downloa
 git add -f node_modules/left_pad/
 ```
 
-Now you can edit the package, commit, and push, and your teammates will get your version of the package. To ensure that your package doesn't get overwritten during an `npm update`, change the default [caret version range](https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004) in your `package.json` to an exact version. Before:
+Now you can edit the package, commit, and push, and your teammates will get your version of the package. To ensure that your package doesn't get overwritten during an `npm update`, change the default [caret version range](https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004) in your `package.json` to an exact version.
+
+Before:
 
 ```json
 "left-pad": "^1.0.2",
--- a/content/using-packages.md
+++ b/content/using-packages.md
@@ -251,7 +251,7 @@ Now `left-pad` is included in your `package.json`, and the code has been downloa
 git add -f node_modules/left_pad/
 ```
 
-Now you can edit the package, commit, and push, and your teammates will get your version of the package. To ensure that your package doesn't get overwritten during an `npm update`, change the default caret version range in your `package.json` to an exact version. Before:
+Now you can edit the package, commit, and push, and your teammates will get your version of the package. To ensure that your package doesn't get overwritten during an `npm update`, change the default [caret version range](https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004) in your `package.json` to an exact version. Before:
 
 ```json
 "left-pad": "^1.0.2",
--- a/content/using-packages.md
+++ b/content/using-packages.md
@@ -233,15 +233,47 @@ sendTextMessage() {
 }
 ```
 
-<h2 id="atmosphere-overriding">Overriding packages from Atmosphere with a local version</h2>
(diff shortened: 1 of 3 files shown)


content/using-packages.md
```diff
@@ -233,15 +233,47 @@ sendTextMessage() {
...
+meteor npm install --save left-pad
+```
+
+Now `left-pad` is included in your `package.json`, and the code has been downloaded to `node_modules/left_pad/`. Add the new directory to source control with:
+
+```bash
+git add -f node_modules/left_pad/
+```
+
+Now you can edit the package, commit, and push, and your teammates will get your version of the package. To ensure that your package doesn't get overwritten during an `npm update`, change the default caret version range in your `package.json` to an exact version. Before:
```
lorensr: Should the package name be changed instead?

tmeasday: Alternatively the package could be removed completely from the `package.json` (although that won't help with dependencies)

tmeasday: Another option is to set it to a path. Maybe this is better?

https://docs.npmjs.com/files/package.json#local-paths

content/writing-packages.md
```diff
@@ -19,11 +20,72 @@ With the release of version 1.3, Meteor has full support for npm. In the future,
...
+
+myPackageLog(); // > "logged from my-package"
+```
+
+<h3 id="including-in-app">Including in your app</h3>
+
+When you are developing a new npm package for your app, there are a couple ways to include the package in your app:
+
+1. Place the package in your app's `node_modules/` directory, and add the package to source control. Do this when you want everything in a single repository.
+1. Place the package outside your app's directory in a separate repository and use `npm link`. Do this when you want to use the package in multiple apps.
```
tmeasday: Can we link to the docs for `npm link`? Also I think it's important to mention this won't go out to others developing your app.

content/writing-packages.md
```diff
@@ -19,11 +20,72 @@ With the release of version 1.3, Meteor has full support for npm. In the future,
...
+
+Then edit the `dependencies` object of `my-app/package.json`, adding `"my-package": "1.0.0"` (use the same version number you chose during `meteor npm init`).
+
+<h3 id="publishing-npm">Publishing your package</h3>
+
+You can share your package with others by publishing it to the npm registry. While most packages are public, you can control who may view and use your package with [private modules](https://docs.npmjs.com/private-modules/intro)).
+
+To publish publicly, [follow these instructions](https://docs.npmjs.com/getting-started/publishing-npm-packages). When you're done, anyone can add your package to their app with `npm install --save your-package`.
+
+If you want to share packages during development, we recommend using source control instead of the registry. If you use the registry, then every time you change the package, you need to increment the version number, publish, and then `npm update my-package` inside your app.
```
tmeasday: Maybe link to https://docs.npmjs.com/files/package.json#git-urls-as-dependencies on [source control]?

tmeasday:
Thanks a lot @lorensr! I think this is useful stuff.

I made a few comments. I do just wonder if the best practice is to commit things in `node_modules/` or a `file://` dependency to another directory.

lorensr:
Thanks! That would be better for avoiding `npm` overwriting, but where should they go? `packages/`? In which case the dependency would be `"my-package": "file:packages/my-package"`

tmeasday:
Oh, yeah, putting in another directory is a bad idea because it'll stuff up node's loading (like you'll need to `require('path/to/package')` rather than `require('package')`). Scratch that.

content/writing-packages.md
lorensr: I was actually referring to the two inclusion methods explained above. Will clarify

lorensr:
Ah, got it.

Finished changes.

tmeasday:
Great work!

stubailo:
Awesome contribution!

lorensr:
Thanks!

PR #383, opened 2016-04-16
Switch order of code style and application structure sections

### LINKED ISSUE #370: Should "Application structure" section follow "Code style"

skirunman:
I think this flows better as it makes more sense to cover code style, etc. before talking about Meteor specific application structure. Resolves #370.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ order: 1001
 description: A log of significant changes to the Meteor Guide.
 ---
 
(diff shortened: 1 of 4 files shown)


tmeasday:
I like this. @stubailo if you agree, can you hit the merge button.

stubailo:
Hit it!

PR #386, opened 2016-04-17
FIxing denormalizer example

d-gubert:
In denormalization example, line 297, "super(doc, callback)" should be "super.insert(doc, callback)".


stubailo:
Thanks!

PR #390, opened 2016-04-20
Changed reference from Atmosphere to Meteor

jamielob:


tmeasday:
Hey @jamielob, thanks. We intentionally refer to them as "Atmosphere packages" to avoid confusion about our support for npm packages.

jamielob:
Hi @tmeasday - Sorry didn't get that from looking at it initially.  So is the entire `Adding files and assets` section specific to Atmosphere?  If that's the case then visually I would recommend that it is either moved underneath or merged with the `Creating an Atmosphere Package` section so that it's clearer.

tmeasday:
Hmm, that's a good point @jamielob. 

Everything below "Creating an Atmosphere Package" is specific to atmosphere and doesn't apply to npm.

I'm not sure what the best approach here is as having 90% of the article under a single heading is kind of silly. Do you think it would have helped if it just said something like "The rest of the content of this article is specific to Atmosphere packages" in the npm package section?

jamielob:
I probably would have missed that sentence.  I'm not sure if you're open to this, but since there seem to be a lot of NPM vs Atmosphere specific information in both the `Using Packages` and `Writing Packages` sections I would re-work the grouping of the Build section to something like below.  Giving a clearer route to deprecating the Atmosphere information and improving the npm. 

```
BUILD

-Atmosphere vs. npm

-Using Atmosphere Packages
--Installing Atmosphere Packages
--Using Atmosphere Packages
--Peer npm Dependencies
--Overriding with a local version

-Writing Atmosphere Packages
--Creating an Atmosphere Package
--Adding files and assets
--Exporting
--Dependencies
--Cordova plugins
--Testing packages
--Local vs. published packages

-Using npm Packages
--Installing npm Packages
--Using npm Packages
--Using Shrinkpack

-Writing npm Packages
--Including in your app
--Publishing your package
```

jamielob:
P.s. Happy to help in any way I can with this @tmeasday

tmeasday:
Hey @jamielob - sounds great!

Sure, send a PR and I'll accept it! 

Reminder to self: ensure redirects work.

PR #392, opened 2016-04-21
Specify the full "dburles:factory" in these example imports

arkadyan:
Apologies if I'm missing something about this, but it seems like these example import lines are incorrectly specifying the factory package.

### DIFF
--- a/content/testing.md
+++ b/content/testing.md
@@ -223,7 +223,7 @@ A simple example of a reusable component to test is the `Todos_item` template. H
 /* eslint-env mocha */
 /* eslint-disable func-names, prefer-arrow-callback */
 
-import { Factory } from 'meteor/factory';
+import { Factory } from 'meteor/dburles:factory';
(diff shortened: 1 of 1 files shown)


tmeasday:
Hi @arkadyan, thanks for the PR.

We actually use a modified version of `dburles:factory` in todos right now: https://github.com/meteor/todos/blob/master/imports/ui/components/client/todos-item.tests.js

You can read a little about what that's all about here: https://github.com/meteor/todos/issues/119#issuecomment-213031559

I'm just not sure the rest of the code will work unmodified with `dburles:factory`. If it does, I think we should change it.

tmeasday:
(But lets get a PR against todos first to prove it).

arkadyan:
@tmeasday Makes sense, thanks for the explanation. It's working for me, but I'm not using `publication-collector` so I'll give it a try in todos and see if I can get anywhere.

PR #396, opened 2016-04-24
Expand CDN section, fixes #198

### LINKED ISSUE #198: Deployment > CDN expansion
(1). I think many people think of a CDN as a thing you can upload static assets to, and the js/css bundles do not feel like static assets, since at least in dev they're always changing filenames, and even if they only changed once per deploy in production, it would be a chore to upload them to the CDN every time. So for those, they might dismiss the idea of CDN, or not understand how it will work with `WebAppInternals.setBundledJsCssPrefix(DNS_HOSTNAME)`. 

So I think it would be helpful to explain 'origin' support, and how a URL is written and fetched from the server if not already cached, as explained in:

http://joshowens.me/using-a-cdn-with-your-production-meteor-app/

cc @queso 

(2). "If you are following the above approach, you may also want to manually add the CDN’s hostname whenever you put an image/other asset URL in your application’s code. To do this throughout your app, you can write a generic helper like imageUrl()"

I think it would be good to clarify that you mean if you have `public/images/foo.png`, then to reference it, do:

```
<img src="{imageUrl "images/foo.png"}">

Template.registerHelper 'imageUrl', (file) ->
  'https://d1477spkfkmftw.cloudfront.net/' + file
```

lorensr:
#198

### DIFF
--- a/content/deployment.md
+++ b/content/deployment.md
@@ -67,7 +67,9 @@ The basic way to use a CDN is to upload your files to the CDN and change your UR
 
 For Meteor, we recommend using a CDN with "origin" support (like [CloudFront](http://joshowens.me/using-a-cdn-with-your-production-meteor-app/)), which means that instead of uploading your files in advance, the CDN automatically fetches them from your server. You put your files in `public/` (in this case `public/cats.gif`), and when your Australian user asks the CDN for `http://mycdn.com/cats.gif`, the CDN, behind the scenes, fetches `http://myapp.com/cats.gif` and then delivers it to the user. While this is slightly slower than getting `http://myapp.com/cats.gif` directly, it only happens one time, because the CDN saves the file, and all subsequent Australians who ask for the file get it quickly.
 
-To get Meteor to use the CDN for your Javascript and CSS bundles, call `WebAppInternals.setBundledJsCssPrefix("mycdn.com")` on the server. This will also take care of relative image URLs inside your CSS files. For all your other files in `public/`, change their URLs to point at the CDN. You can use a helper like `assetUrl`.
+To get Meteor to use the CDN for your Javascript and CSS bundles, call `WebAppInternals.setBundledJsCssPrefix("mycdn.com")` on the server. This will also take care of relative image URLs inside your CSS files. If you need to use a dynamic prefix, you can return the prefix from a function passed to `WebAppInternals.setBundledJsCssUrlRewriteHook()`.
+
+For all your files in `public/`, change their URLs to point at the CDN. You can use a helper like `assetUrl`.
 
 Before:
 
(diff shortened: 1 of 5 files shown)


lorensr:
@martijnwalraven Hi, would you please check over my section on Cordova assets? Also, is there a way to tell your native code to fetch things from a CDN instead of the meteor server? Is the websocket the only thing that needs to go to the meteor server, or does the index need to come from there as well? If can get index from CDN, then could do something like `meteor build --app-server http://mycdn.com --ddp http://myapp.com`

tmeasday:
@lorensr non-mobile bit looks great! I'm not sure if the recommendation for mobile is correct, I'll let @martijnwalraven take it from here.

content/deployment.md
```diff
@@ -61,11 +61,39 @@ It's always a good idea to use SSL for Meteor applications (see the [Security Ar
...
+});
+```
+
+```html
+<img src="{{assetUrl "cats.gif"}}">
+```
+
+<h4 id="cordova-assets">Cordova assets</h4>
+
+In Meteor versions 1.3 and up, if your app is mainly meant to be used inside Cordova, we recommend **not** using a CDN. Meteor Cordova caches your `myapp.com/*` files on the phone, and it's much faster to get the phone copies than request them from a CDN.
```
martijnwalraven: This bit may be a little confusing, as `WebAppInternals.setBundledJsCssPrefix` (or `setBundledJsCssUrlRewriteHook()`) doesn't actually affect Cordova. So there is no reason not to use these for your web app even if you also build a mobile app from the same codebase.

The takeaway message is that Cordova will always load assets locally, unless you use an absolute URL to manually load them from a CDN (or other server). You wouldn't want to do that for assets that are required to initially load your app, but there is a tradeoff between having assets bundled as part of your app (requiring them to be downloaded from the app store and stored on your device) and loading them on demand.

content/deployment.md
```diff
@@ -61,11 +61,39 @@ It's always a good idea to use SSL for Meteor applications (see the [Security Ar
...
+
+Pre-1.3 Meteor Cordova still cached your files on the phone, but it re-downloaded *all* the files during each hot code push, and it did so from main-thread Javascript, which froze the UI. In order to shorten the duration of the UI freeze, you had to remove the files from `public/` and upload them to a CDN.
```
martijnwalraven: I'd leave this out, because the mobile guide is only meant for 1.3. Also, I would advise anyone still on a pre-1.3 version of Meteor to upgrade rather than try to work around these issues by uploading all your assets to a CDN (which has other downsides, because now the assets are no longer available locally).

lorensr:
Sorry, in future will not clutter up log 😣

Thanks @martijnwalraven !

stubailo:
@lorensr use the "squash and merge" button!

lorensr:
There were conflicts, so no button on this one, but will do

PR #404, opened 2016-04-27
Quiet eslint errors from newly created app

willxy:
When following the guide, I wasn't expecting eslint to give errors from a newly created default app; it did; I found some rules to quiet them; Sashko suggested a PR to note them.  I worded them as "you can" rather than "you should" as I'm not familiar enough to know what's recommended.


stubailo:
Great improvement!

PR #412, opened 2016-04-30
Clarify methods of importing styles

lorensr:
Added comments at the head of snippets with a filename, so the reader knows the extension. Was a confusing mixture of less, js, and css. 

Instead of `foo.js/less`, could do `imports.` or `styles.`, `my-component.`, or full path, `client/imports.`

Removed the example of importing npm css from css, because it appears not to work (https://github.com/meteor/meteor/issues/6846)

### DIFF
--- a/content/build-tool.md
+++ b/content/build-tool.md
@@ -1,5 +1,5 @@
 ---
-title: Build system
+title: Build System
 order: 33
 description: How to use Meteor's build system to compile your app.
 discourseTopicId: 19669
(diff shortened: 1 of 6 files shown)


stubailo:
Hmm, the code snippets already have the language in the top right, maybe we can just make that more prominent in the css?

content/build-tool.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Build system
+title: Build System
```
skirunman: Can we get agreement on style here. We have been making all titles only first word capitalized in <h> tag. Now folks are wanting to change it back. I personally don't care, but it would be best to stay with one approach and everything for the most part is only capitalizing the first word.

lorensr: Haha, I was following data/view/build/production titles ☺️ I'd prefer lowercase. @stubailo ?

lorensr:
> Hmm, the code snippets already have the language in the top right, maybe we can just make that more prominent in the css?

No language marking for less atm, but now that I look, think it's just due to a missing css `:after` style. 

I think it would have to be a pretty significant style to make sure people look at it when reading through the text (like large and red, and/or on the top left), but the current muted styling makes sense for most of the Guide, where the filetype isn't important – it's just there on side if you wanted it. Style imports is the one place where it's a surprise & important for understanding.

image

content/build-tool.md
stubailo: I think title should be title case, but headers should be sentence case. That seems consistent with style I see elsewhere, but we could just use sentence case everywhere!

stubailo: Also once we agree let's add it to the contribution guidelines

stubailo:
Hmm I mean if we feel like we need to start putting language indicators in comments, that's a sign that the default ones should be more prominent IMO

stubailo:
Perhaps let's remove that from this PR and open a new issue about how to emphasize that stuff? Also let's get Dom in on this on Monday, he's actually redesigning the guide theme this week (plus docs etc)

tmeasday:
We should probably add a LESS (and anything else we want) class to https://github.com/meteor/hexo-theme-meteor/blob/master/source/style/content.styl#L113-L136

PR #419, opened 2016-05-05
Clarify using ROOT_URL vs. --mobile-server

awwx:


tmeasday:
Thanks @awwx. Work for you @martijnwalraven ?

martijnwalraven:
I don't think this is actually true. For `meteor run`, we use `--mobile-server` to set `ROOT_URL` in `__meteor_runtime_config__` and to set default access rules for Cordova's `config.xml`. (For `meteor build`, `--server` serves the same purpose).

Setting the `ROOT_URL` environment variable is something you have to do additionally on the server, which sets it on `__meteor_runtime_config__` in the generated boilerplate for hot code push.

awwx:
Ah, I had missed the automatic whitelisting part.

PR #423, opened 2016-05-07
Add filenames before test snippets?

lorensr:
😏 I know what you're thinking... *not filenames again* 😒

I think it's helpful to see whether it's a `.test.js` or an `.app-test.js`. I also think it's helpful to see the file structure, as a suggestion to those (like me) reading that section and wondering where to put the test I'm about to write.


stubailo:
Good idea!

PR #440, opened 2016-05-12
Update pr template

stubailo:
I'm tired of seeing the boilerplate text!

One thing that worked well for apollo-client was adding a checklist in the body: https://github.com/apollostack/apollo-client/blob/master/.github/PULL_REQUEST_TEMPLATE.md

What would we put in such a checklist?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,7 +1,9 @@
 <!--
-  Thanks for submitting a PR! We'll try to look at it as soon as possible.
-
-  Use this field to describe your pull request.
-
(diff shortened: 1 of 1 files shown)


lorensr:
I agree, always takes me a second to be like what's going on, then
remember, then look down 😄
Either "If you are adding significant new content, please take a moment to
include an update to the changelog
<https://github.com/meteor/guide/blob/master/CHANGELOG.md> in your PR." or
nothing?
Mental tiring of seeing any kind of boilerplate at all versus having to
change changelog ourselves a little more often?
I'd go with nothing, but happy w wtv

stubailo:
Curious what @tmeasday thinks

tmeasday:
A change log checklist seems like a good start

lorensr:
Suggested changes in commits ^

stubailo:
Thanks!

PR #443, opened 2016-05-13
Apollo article 🚀

lorensr:
TODO:
- [x] section in apollo docs on mutators that we can link to?
- [ ] sashko fact-checking me, i just guessed sometimes 😊
- [ ] advanced section
- [ ] hexo, including discourse topic
- [ ] meteor needs its own emoji if it wants to keep its brand in the 21st century.
1. 💫
2. 🌠
3. ☄️
4. 🌌


stubailo:
Whoa, intense! ill take a detailed look tomorrow!

lorensr:
> Whoa, intense!

Yeah I know, I can't decide either. Torn between 2 and 4 😕

lorensr:
For the advanced section: the server user lookup in meteor integration - context isn't saved, right? For each query you look it up, then provide it to resolvers? So added latency in comparison to Meteor. Plans in future to have sessions / memory cache on app servers?

stubailo:
> For each query you look it up, then provide it to resolvers? So added latency in comparison to Meteor. 

Yes, that's true - in Meteor the `userId` is saved on the connection. But if you ever want to get the user object itself, you have to query it every time. So I don't know if it's a big deal, but it would be nice to have some sort of caching for the user ID in particular.

Actually, for a particular login token the user ID will never change, right? So it could be safe to just have a global cache dictionary of login token -> user ID.

lorensr:
> user ID will never change, right?

Right. I can add the dictionary to the integration example. Will Apollo
have an accounts system (and sticky routing I guess) that the dictionary
would be part of?

stubailo:
Eventually we are going to move all of Meteor accounts in there, but probably not in the next month or two.

One way to take advantage of sticky routing could be to do GraphQL-over-DDP, but that means you can't use express-graphql anymore, which would be a hassle. I'd say it's not worth it for now.

tmeasday:
I'm confused by this, wouldn't most load balancers (galaxy included) do
sticky sessions at the http level too?

stubailo:
Oh yeah, fair point. So it's not a concern at all. Shows what I know about HTTP :P

StorytellerCZ:
I guess this needs to be updated.

StorytellerCZ:
I have updated this article for now. I think we should merge this as is and then we can iterate on it in the future.

content/apollo.md
```diff
@@ -0,0 +1,102 @@
+---
+title: Apollo
+order: 15
+description: The Apollo data stack for Reactive GraphQL
+discourseTopicId: TODO
+---
+
+<h2 id="introduction">Introduction</h2>
+
+Apollo is Meteor's next-generation data stack, built with GraphQL. While it doesn't yet have all the features that Meteor's pub/sub system has, it provides a way to get data from any database – not just MongoDB.
```
filipenevola, line 10: > Apollo is Meteor's next-generation data stack 

I don't believe this open statement is correct.

Apollo is not part of Meteor or a next generation of Meteor

content/apollo.md
```diff
@@ -0,0 +1,102 @@
...
+
+<h3 id="client">Apollo Client</h3>
+
+[Apollo client docs](https://www.apollographql.com/docs/react/)
+
+<h4 id="getting-data">Getting data</h4>
+
+Instead of calling `Meteor.subscribe`, you will use [queries](https://www.apollographql.com/docs/react/data/queries/) to get data.
+
+The main different with subscriptions is that queries get called only once (by default) and don't get updated data like a subscription would. This is great for data that doesn't change often and where you don't need reactivity.
```
filipenevola, line 27: > different 

difference

content/apollo.md
```diff
@@ -0,0 +1,102 @@
...
+- [Intro to GraphQL](https://medium.com/apollo-stack/the-basics-of-graphql-in-5-links-9e1dc4cac055)
+- [GraphQL coming from REST](https://medium.com/apollo-stack/how-do-i-graphql-2fcabfc94a01#.pfdj5bxxj)
+
+<h3 id="advanced">Advanced<h3>
+
+[Principled GraphQL](https://principledgraphql.com/)
+
+<h4 id="latency">Latency</h4>
+
+Meteor publications are blocking, whereas multiple GraphQL queries are executed in parallel. Publications stream data to the client as it arrives, whereas all the resolvers in a GraphQL query have to return before the data is sent to the client. (Although GraphQL is discussing adding the ability to stream results to the client as they come in.)
```
filipenevola, line 63: > Meteor publications are blocking

Meteor publications are blocking by default

content/apollo.md
```diff
@@ -0,0 +1,102 @@
...
+  Query: {
+    userUniverses: async (obj, { hideOrgs }, { user }) => {
+      if (!user) return null
+      const selector = { userId: user._id, }
+      if (hideOrgs) selector.organizationId = { $exists: false }
+      return UniversesCollection.find(selector).fetch()
+    }
+  }
+}
+```
```
filipenevola, line 102: It would be helpful to list other packages, like https://github.com/quavedev/graphql that provides more features.

filipenevola (changes requested):
I've suggested a few changes.

Also, are the changes on site/themes intentional?

content/apollo.md
StorytellerCZ: Added, but it would be nice if it was updated for the latest version of Apollo.

lorensr:
Wow, from over 5 years ago! Thanks @StorytellerCZ for updating ☺️

PR #447, opened 2016-05-16
Added a missing word

johnnynotsolucky:
"tools that you can to specify" -> "tools that you can use to specify"


lorensr:
Thanks 🙌  😊

PR #448, opened 2016-05-17
Document Meteor.isAppTest

benweissmann:
The testing article doesn't mention that full-app test mode sets `Meteor.isAppTest` instead of `Meteor.isTest`. This PR just adds a short note to the list of differences between full-app test mode and unit test mode.


stubailo:
I thought it set both flags? Or just `isAppTest`?

tmeasday:
Nope, just `isAppTest`

PR #455, opened 2016-05-19
fixed boolean logic priority bug

qbolec:

There is a subtle difference between `(! a === b)` and `(a !== b)`.
For example `(! 2 === 3)` is false, while `(2 !== 3)` is true.


stubailo:
Good catch!

PR #459, opened 2016-05-22
Update methods.md

zanot:
In templates events handler, current instance of the template can be accessed in the parameters of the handler.


lorensr:
Thanks 😊

PR #465, opened 2016-05-27
Fixing `export const function` syntax error

tjconsult:
`export const function` results in a syntax error. Fixing one example by just dropping it and changing to a default export for the template test helper.

### DIFF
--- a/content/testing.md
+++ b/content/testing.md
@@ -223,7 +223,7 @@ const withDiv = function withDiv(callback) {
   }
 };
 
-export const function withRenderedTemplate(template, data, callback) {
+export default function withRenderedTemplate(template, data, callback) {
   withDiv((el) => {
     const ourTemplate = _.isString(template) ? Template[template] : template;
     Blaze.renderWithData(ourTemplate, data, el);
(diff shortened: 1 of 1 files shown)


tmeasday:
Hi @tjconsult - I think maybe we should change it to fit what we currently have in the todos app? https://github.com/meteor/todos/blob/master/imports/ui/test-helpers.js#L18

tjconsult:
Agreed–changes made. Mind if I am being nitpicky and drop the second superfluous `withRenderedTemplate`? I could update the TODOs app example, too, if you want me to, unless there are side-effects I am not aware of.

tmeasday:
It may not pass linting? I'd love you to update Todos and check there. Let's do that first.

tjconsult:
@tmeasday – you're right, [making that change would cause lint to fail](https://github.com/meteor/todos/commit/b2c540a191199736c36362342f027e2df99e2061#diff-eaee00df3b1612f61cd59924794db3feL18) - you put it in there to make AirBNB's rules pass. I tried a couple alternatives but there's a reason both the lint rule exists _and_ the export is `const`, so I guess I'll update the guide to match the TODOs example code.

tmeasday:
Damn I was hoping there was a better solution. Thanks for looking into it :)

PR #471, opened 2016-05-31
Add link to PascalCase in c2 wiki

icortex:
TODO:


lorensr:
Hey Victor, thanks! I'd guess https://en.wikipedia.org/wiki/PascalCase would be a more reliable place to point to?

icortex:
@lorensr You are right. Updated!

Issue #470, opened 2016-05-31
[cordova] Explain that device permissions must be requested at run-time


abernix:
Per meteor/meteor#6976, as well as:
- https://forums.meteor.com/t/resolved-android-marshmallow-runtime-permissions-for-microphone-and-local-storage/22149/4
- https://github.com/dpa99c/cordova-diagnostic-plugin/#android-runtime-permissions

It needs to be explained to people that they need to request permissions to use camera, microphone, filesystem, etc. at run-time.

(This is just what I'm gathering – I unfortunately do not know Cordova very well/at all, but I will try to get back to this eventually with a PR.  Just want to leave this here before it gets forgotten.)

abernix:
Per @lorensr in meteor/meteor#6976:

@inspiraluna here are the three places to make note:

https://github.com/meteor/guide/blob/master/content/mobile.md

(I guess in the configuring-your-app section, saying how to do Android permissions in pre-6 and in 6)

https://github.com/meteor/mobile-packages/blob/master/packages/mdg:camera/README.md
https://github.com/meteor/mobile-packages/blob/master/packages/mdg:geolocation/README.md

inspiraluna:
okey... I did it this morning! Hope its not too bad. ;)
please check...

abernix:
This was fixed in #472 thanks to the great contribution by, @inspiraluna.  Thanks!

(Closing)

PR #472, opened 2016-06-01
Update mobile.md

inspiraluna:


abernix:
This is in reference to #470.

abernix:
I'll start my review with the warning that I don't understand or use Cordova! :wink:

[This page on Google](https://developer.android.com/training/permissions/requesting.html) makes it sound like they need to be listed in the app manifest too?  Is this true?  Or is it only at runtime?

I'll make additional notes as line comments.

content/mobile.md
```diff
@@ -452,6 +452,33 @@ What is often confusing to people is that setting `App.accessRule` is not enough
 
 To get around these restrictions, you'll have to use what is known as [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS). In contrast to the whitelisting mechanism configured on the client, CORS relies on headers set by the server. In other words, in order to allow access to a remote resource, you may have to make configuration changes on the server, such as setting a `Access-Control-Allow-Origin` header.
 
+<h3 id="csp">Runtime Permissions</h3>
```
abernix: I might say `System Permission` since that is what Google calls it?

https://developer.android.com/training/permissions/requesting.html

content/mobile.md
```diff
@@ -452,6 +452,33 @@ What is often confusing to people is that setting `App.accessRule` is not enough
...
+When using e.g. camera, microphone or any other native mobile phone features in Cordova, since Android 6 (Marshmallow) it became necessary to request permissions additionally at runtime. In such case you need to add the cordova diagnostic plugin which manages exactly this. (https://github.com/dpa99c/cordova-diagnostic-plugin)
```
abernix: I would word this as:

As of Android Marshmellow, certain system features such as the camera, microphone, etc. require additional permissions in order to access them.  These must be listed in the manifest and also requested at runtime.

(Specify here if how to add them to the manifest if that's necessary???)

To request them at runtime, you will need to use the the `[cordova.plugins.diagnostic](https://github.com/dpa99c/cordova-diagnostic-plugin)` plugin.

content/mobile.md
```diff
@@ -452,6 +452,33 @@ What is often confusing to people is that setting `App.accessRule` is not enough
...
+
+For example, here we prepare our android and ios hardware permissions for a WebRTC session
```
abernix: Needs proper capitalization on Android and iOS.

abernix:
Thanks for doing this @inspiraluna.  Overall it looks great.  I made a few comments, and I'll leave this for further review from the docs team, and probably @martijnwalraven.

content/mobile.md
inspiraluna: not sure if this is called systems permissions or runtime permission or just permissions on Android and it needs to be checked if its called differently on iOS too. I used the term from here: https://github.com/dpa99c/cordova-diagnostic-plugin#android-runtime-permissions

content/mobile.md
inspiraluna: I'm German - thanks for correcting this with proper English ;)
And yes its right it must be listed in manifest (for Android) too!

lorensr:
@inspiraluna Sorry for the delay, it's looking good, thank you. 

"you will need to use the the [cordova.plugins.diagnostic]"

I don't believe this package is *necessary* – it's just helpful, right? You can just attempt to access eg camera, and the first time you do, the user will be prompted for permission. Let's say they say yes. Then:

- on iOS, you're never prompted for permissions again. 
- on Android 6, what happens? I would guess the same.

abernix:
I've made some slight changes to the suggested text but in an effort to get this PR merged, I conceded that it may not be _necessary_ to use `cordova.plugins.dignostic` but may be recommended.

My changes are in:
* aebca1b4e83b5eed5ab710ef597036288cc45d73
* 2432bf65722c7bc720d377f0b8260cf5f717f104
* 72f17e433b21295dc5e8d4ab70203e6355f9ae9b

Happy to tweak more if necessary! :)

PR #478, opened 2016-06-08
routing.md: fix `this` in Template...onCreated example

abesto:
The ES6 `=>` operator binds the `this` context inside the function to whatever it is where the function is defined. In this case, that'll make `this` inside the `onCreated` handler the `window` object, instead of the intended `TemplateInstance` object.


tmeasday:
Thanks @abesto, this is good. Can we take this first as a PR to https://github.com/meteor/todos, so we can verify it works (in this case, I suspect the above code will need to be tweaked a little to pass linting)?

abesto:
There's already code like this in the TODOs app where `this` is used inside the handler, like in https://github.com/meteor/todos/blob/82b493da1893d8e5d6bca60f212cdab841f31903/imports/ui/components/todos-item.js#L17. I've added a commit that makes the example here look just like the already existing code in TODOs.

tmeasday:
Oh, great, thankyou!

PR #479, opened 2016-06-08
removed redundant if

qbolec:
TODO:

I removed an if which I believe was redundant. In case it was really necessary, and I missed it, then I would suggest to add a comment with explanation for it's existence.

Also, I'm not sure what (if any) exceptions can be thrown by `this.added(...)`, but perhaps it would be a better idea to move `publishedKeys[doc._id]=true` after the `this.added(...)`?


tmeasday:
Thanks @qbolec !

PR #481, opened 2016-06-10
Change default to const to avoid linter errors.

hwillson:
Hi guys - see [this forum post for the issue](https://forums.meteor.com/t/meteor-guide-methods/19662/31). Changing `default` to `const` will help address this. Thanks!


tmeasday:
Thanks!

PR #494, opened 2016-06-29
Update collections.md with reference to Astronomy.

JesperWe:
Simple-schema is not the only game in town ;-)


tmeasday:
Sure, let's have it.

JesperWe:
Wow. That was the quickest PR -> Merge so far, I think: 4 minutes! :+1:

PR #495, opened 2016-06-29
Removed Reapp recommendation

JWPapi:
TODO:

Reapp is not maintained good, I would not recommend it. Even there demo is off for 30 days

https://github.com/reapp/reapp/issues/119


lorensr:
Thanks! Looks like the other main option, Touchstone, is similarly unmaintained https://github.com/touchstonejs/touchstonejs/issues/120

dnish:
I think https://onsen.io/v2/docs/guide/react/ could be a recommendation.

lorensr:
Thanks, I'll add that – similar number of stars, actively developed, demo works well, styling for both iOS and Android components, and docs are great.

PR #496, opened 2016-06-29
Translation es: Index

hedaro:
I translated the Index.md file, which are for the 1.2 versión.
I tried to translate all technical terms the best way i could, but i think some could be improved, like **client connected**. There are a few terms i left untranslated because i don't know any technical equivalent in spanish. I'll add them if anyone knows the right translation.
Some of these terms are:
- Full stack
- Data on the wire
- Pull request

I left the terms **issues** and **docs** untranslated because they refer to sections of a site (github issues and meteor docs).
I tested all texts with a spell checker, so orthography shouldn't be an issue. Still any corrections and grammar suggestions are appreciated.

### DIFF
--- a/content/index.md
+++ b/content/index.md
@@ -16,13 +16,13 @@ Meteor es una plataforma full-stack de JavaScript para desarrollar aplicaciones 
 
 - Meteor provee **reactividad full stack**, permitiendo a la interfáz gráfica reflejar sin problemas el verdadero estado del mundo con mínimo esfuerzo de desarrollo.
 
-<h2 id="what-is-it">What is the Meteor Guide?</h2>
+<h2 id="what-is-it">¿Qué es la Guía de Meteor?</h2>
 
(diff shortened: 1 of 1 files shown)


tmeasday:
Thanks @hedaro! Can you merge current `master` into your branch, and we'll get a deployment going?

hedaro:
@tmeasday done.

tmeasday:
Hmm, I think maybe you'll need to send another PR, I can't seem to merge that.

hedaro:
I'm gonna be watching #499, because i did the same thing for the spanish branch, so if that's not the way to go then we first need to figure things out.
It'd help to know exactly what is expected from our branches/forks to get the deployment to work.

tmeasday:
Hey @hedaro, basically we just need the changes to `circle.yml` that are on `master`. But it's probably good to merge `master` into your branch every now and then to avoid the untranslated files going out of date.

PR #498, opened 2016-06-30
Edited section on React data to use best practices

ffxsam:
TODO:

If a component is called XYZ, it's usually wise to name the container
XYZContainer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ order: 1001
 description: A log of significant changes to the Meteor Guide.
 ---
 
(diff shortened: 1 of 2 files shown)


tmeasday:
We should also probably update the ui/ux article use the "container"/"presentational" taxonomy as that seems to have been what people have settled on.

@ffxsam if you are inclined, it'd be great to update the react branch of todos to reflect this.

ffxsam:
@tmeasday Sure! I can actually think of a whole bunch of things to modify in the React Todos, if you don't mind.

tmeasday:
Sure, go for it!

PR #504, opened 2016-07-02
Add guidance on i18n with React

JeremySaks:
This adds some guidance on i18n with React using the [`universe:i18n`](https://github.com/vazco/meteor-universe-i18n) package. There have been [issues](https://github.com/TAPevents/tap-i18n/issues/115) for a while now using `tap:i18n` in React. @stubailo [suggested](https://forums.meteor.com/t/does-meteor-has-offical-ways-to-handle-with-translation-like-i18n-in-ruby-on-rails/14235/8) addressing this in the guides so I went ahead!

Disclaimer: I'm not a `universe:i18n` expert, just a happy customer. But it's usage seems to be increasing. I've tested the guidance in this pr with [`simple-todos-react`](https://github.com/meteor/simple-todos-react) without issue. Changelog is updated.


tmeasday:
Looks great! Thanks @JeremySaks

PR #505, opened 2016-07-04
Update instructions for juliancwirko:postcss

detrohutt:
...to reflect changes in the meteor/todos app ( https://github.com/meteor/todos/commit/a127e3128fca6eee98a2476896243677a17a9adf )

Not sure if this constitutes a major change. I'd think not, but I suppose if people have been installing autoprefixer as a devDependency and are having issues, it could help to draw attention to the change. I also considered adding a note above the instructions such as:

"Note: a previous version of the guide mistakenly suggested adding autoprefixer as a devDependency. If you used a previous version of the guide and are having issues, please ensure autoprefixer is installed as a normal dependency(and don't forget to 'npm install')."

I've neither added the note, nor updated the changelog, but I'll be more than happy to update this PR however you see fit.

### DIFF
--- a/content/build-tool.md
+++ b/content/build-tool.md
@@ -173,7 +173,7 @@ Then we can install any npm CSS processing packages that we'd like to use and re
 
 ```
 {
-  "devDependencies": {
+  "dependencies": {
     "autoprefixer": "^6.3.1"
   },
   "postcss": {


abernix:
Thanks for helping in getting this stuff sorted out.  Much appreciated!

Interesting that `autoprefixer` _had_ to be a regular `dependency` instead of a `devDependency` though.  Did you confirm this?

It's strange since all the pre-deploy minification happens in the dev environment during `meteor build` and I wouldn't expect `autoprefixer` to be needed in production.  `meteor build` would need to load `devDependencies` itself in order to run the build plugins properly, but for the actual outputted package, it doesn't seem necessary.

Let me know your thoughts.

tmeasday:
I'm not sure there's any practical difference between the two anyway. `meteor build` doesn't look at `package.json`, it's simply copies modules over from `node_modules` based on a static analysis of `require`s.

detrohutt:
Unfortunately I don't have a deployed app to test with. Everything you're saying makes sense to me. Perhaps @tmeasday could offer some insight since he's the one who changed it from a dev dependency?

detrohutt:
I'll close this PR and continue to suggest using 'devDependencies' until it becomes clear whether or not it's actually required to be 'dependencies'. Thanks for the input guys!

PR #508, opened 2016-07-07
Typo in component name

Xanadhi:
TODO:

I think ListContainer should be ListPageContainer according to the tutorial


lorensr:
Thanks @Xanadhi 

Couple more instances as well. `todos/react` uses `ListContainer`. @tmeasday which is best?

tmeasday:
I think we should do whatever's standard in React-land these days. I think it's `XContainer`.

lorensr:
Thanks, I'll fix the other occurrences and todos

PR #512, opened 2016-07-12
fixes typo

discdiver:
was "make" is now "makes" in "Usually, while developing an application, it make sense to run meteor test on a second port (say 3100), while also running your main application in a separate process:"


lorensr:
Thanks 😊

PR #514, opened 2016-07-14
Fix and harden migration database operations

sebakerckhof:
Downgrades needed `{multi: true}` added and batch needs to contain at least one operation or Mongo will throw an error and migration will be in locked state...

### DIFF
--- a/content/collections.md
+++ b/content/collections.md
@@ -327,7 +327,7 @@ Migrations.add({
     });
   },
   down() {
-    Lists.update({}, {$unset: {todoCount: true}});
+    Lists.update({}, {$unset: {todoCount: true}}, {multi: true});
   }
 });
 ```
@@ -358,12 +358,17 @@ Migrations.add({
       batch.find({_id: list._id}).updateOne({$set: {todoCount}});
     });
 
-    // We need to wrap the async function to get a synchronous API that migrations expects
-    const execute = Meteor.wrapAsync(batch.execute, batch);
-    return execute();
+    //Mongo throws an error if we execute a batch operation without actual operations, e.g. when Lists was empty.
+    if(batch.s && batch.s.currentBatch && batch.s.currentBatch.operations && batch.s.currentBatch.operations.length > 0){
+      // We need to wrap the async function to get a synchronous API that migrations expects
+      const execute = Meteor.wrapAsync(batch.execute, batch);
+      return execute();
+    }
+  
+    return true;
   },
   down() {
-    Lists.update({}, {$unset: {todoCount: true}});
+    Lists.update({}, {$unset: {todoCount: true}}, {multi: true});
   }
 });
 ```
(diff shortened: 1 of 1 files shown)


content/collections.md
```diff
@@ -358,12 +358,17 @@ Migrations.add({
       batch.find({_id: list._id}).updateOne({$set: {todoCount}});
     });
 
-    // We need to wrap the async function to get a synchronous API that migrations expects
-    const execute = Meteor.wrapAsync(batch.execute, batch);
-    return execute();
+    //Mongo throws an error if we execute a batch operation without actual operations, e.g. when Lists was empty.
+    if(batch.s && batch.s.currentBatch && batch.s.currentBatch.operations && batch.s.currentBatch.operations.length > 0){
```
abernix: Wouldn't it be more appropriate (i.e. safer) to use the (documented) [`Bulk.tojson` method](https://docs.mongodb.com/manual/reference/method/Bulk.tojson/#Bulk.tojson) to access this information instead of using the (to my knowledge, undocumented) `s.currentBatch` attributes?

Maybe the information presented isn't quite the same and I don't have a test script immediately available (maybe you do?), but it seems that `batch.tojson()` should return something along the lines of:

```
{ “nInsertOps” : 2, “nUpdateOps” : 0, “nRemoveOps” : 1, “nBatches” : 2 }
```

...where it might even be acceptable to just check if `nBatches > 0` before preceding.

abernix: I'm actually not sure the latest version of Mongo still throws an error when bulk ops are executed without any operations:

```js
rs0:PRIMARY> db.testtesttest.initializeOrderedBulkOp().execute()
BulkWriteResult({
	"writeErrors" : [ ],
	"writeConcernErrors" : [ ],
	"nInserted" : 0,
	"nUpserted" : 0,
	"nMatched" : 0,
	"nModified" : 0,
	"nRemoved" : 0,
	"upserted" : [ ]
})
```

Thoughts?

sebakerckhof:
`batch.toJson()` seems to be a better approach indeed, didn't know about that (actually got the `s.currentBatch` from a stackoverflow answer).

I posted this when I ran into this problem using Meteor 1.2 (Mongo 2.6), so it might not be an issue anymore...

sebakerckhof:
Sounds good to me.
I just wonder what would happen if the find in a statement like this:
`batch.find({_id: list._id}).updateOne({$set: {todoCount}});`

Would match no records. 
Although this can't happen in the case of this example in the guide, imagine you do something like:
```js
...
let hasUpdates = false;
const batch = Todos.rawCollection().initializeUnorderedBulkOp();
Lists.find({}).forEach(list => {
   hasUpdates = true;
    batch.find({listId: list._id}).updateOne({$set: {listTitle:list.title}});
});
...
```
And no list has any todo's yet. 
There are probably other workarounds for that, but the `s.currentBatch.operations.length`, although undocumented, will always be correct.

Anyway, that's an edge case of an edge case, so just using the `hasUpdates

abernix:
> I just wonder what would happen if the find in a statement like this:
> batch.find({_id: list._id}).updateOne({$set: {todoCount}});

The batch won't care so it should work fine.

> although undocumented, will always be correct.

Yeah, just would rather avoid that if not overly complicated.

If you can make those changes, this looks good to me and can be merged.

sebakerckhof:
done

abernix:
Thanks, @sebakerckhof .

PR #517, opened 2016-07-16
Remove duplicate section

raffomania:
TODO:

The page "Using Atmosphere Packages" contained two identical versions of the section "Peer npm dependencies". This commit removes the second instance.


lorensr:
Haha oops, thank you 😊

PR #519, opened 2016-07-17
Update blaze.md

DylanKojiCheslin:
TODO:

Each and With section references each-in section as being below it, incorrectly.


lorensr:
Thanks Dylan 😊

PR #522, opened 2016-07-20
Typo # instead of .

shwaydogg:
Unless the # in Array#forEach is a nomenclature I don't know of!


lorensr:
Thanks Greg 😊

Slava:
`#` is something that denotes an instance method. The use of `#` is intentional throughout the guide and docs. @lorensr maybe rethink this merge?

lorensr:
Oh, thanks! My google search didn't find the #, so I figured the more familiar . was better. Will change back 😊

PR #526, opened 2016-07-27
Added observed gotchas

brucejo75:
TODO:

- [N/A ] If this is a significant change, update [CHANGELOG.md](https://github.com/meteor/guide/blob/master/CHANGELOG.md)
- [X ] Use `<h2 id="foo">` instead of `## Foo` for headers
- [X ] Leave a blank line after each header

Hi @tmeasday, this is a list of gotchas I ran across while making v1.4 work.  @stubailo suggested I add [this ](https://forums.meteor.com/t/meteor-1-4-not-starting-due-to-debugger-statements/27109)to the migration guide, so I thought I would give it a shot and add all the issues I ran into.  Not sure this is a format or the way you would want to address any of this.  You are likely a better editor than I.

Also, I used `<h4>` headers without any IDs in one section...

### DIFF
--- a/content/1.4-migration.md
+++ b/content/1.4-migration.md
@@ -64,3 +64,44 @@ if (Meteor.isClient) {
 ```
 
 One place this is particularly useful is in [test files that are only intended to run on the client or the server](https://github.com/meteor/todos/commit/3963a65d96cd7ef235a95d5e3a331d6f0606f70f) &mdash; you can now use `import` wherever you like, without having to organize your tests in `client` or `server` directories.
+
+<h2 id="gotchas">Gotchas!</h2>
+Here is a list of observed gotchas...
+
+<h3 id="mongo-3_2-gotchas">Mongo 3.2</h3>
+
+If you are using [oplog tailing](https://github.com/meteor/docs/blob/version-NEXT/long-form/oplog-observe-driver.md) and you see a failed authentication you may need to upgrade to [SCRAM-SHA-1](https://docs.mongodb.com/manual/release-notes/3.0-scram/#upgrade-mongodb-cr-to-scram), essentially: `use admin, db.adminCommand({authSchemaUpgrade: 1});`.  You may need to delete and re-add your oplog reader user.
+
+<h3 id="node-gyp-gotchas">node-gyp</h3>
+
+Make sure this is installed so you can compile your node modules to native if necessary.  See [Install `node-gyp`](https://github.com/nodejs/node-gyp#installation)
+
+<h3 id="clear-out-oldstuff-gotchas">Clear out old stuff</h3>
+
+Try a `meteor reset` or better yet `rm -rf .meteor/local`.  Follow this with the [rebuild command](https://guide.meteor.com/1.4-migration.html#binary-packages-require-build-toolchain).
+
+<h3 id="npm-bcrypt-gotchas">npm-bcrypt</h3>
+
+You may run into this [`npm-bcrypt`](http://stackoverflow.com/questions/29340510/bcrypt-is-breaking-my-meteor-application-how-do-i-fix-it/29347616) problem
+
+<h3 id="debugger-gotcha">debugger statement issue</h3>
+If you have `debugger` statements in your code they will now hit the breakpoint even without a debugger attached.
+<h4>Old Behavior (prior to Meteor 1.4 with Node v0.10.46)</h4>
+
+Set `NODE_OPTIONS=--debug`<br>
+Sprinkle `debugger` statements in code<br>
+`meteor` command will start and run fine, skipping `debugger` statements
+
+<h4>New Behavior (Meteor 1.4 with Node v4.4.7)</h4>
+
+Set `NODE_OPTIONS=--debug`<br>
+Sprinkle `debugger` statements in code<br>
+`meteor` command will start will stop at first `debugger` statement waiting for a debugger to attach
+
+<h4>Comments:</h4>
+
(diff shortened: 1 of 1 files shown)


content/1.4-migration.md
```diff
@@ -64,3 +64,44 @@ if (Meteor.isClient) {
...
+<h2 id="gotchas">Gotchas!</h2>
+Here is a list of observed gotchas...
+
+<h3 id="mongo-3_2-gotchas">Mongo 3.2</h3>
+
+If you are using [oplog tailing](https://github.com/meteor/docs/blob/version-NEXT/long-form/oplog-observe-driver.md) and you see a failed authentication you may need to upgrade to [SCRAM-SHA-1](https://docs.mongodb.com/manual/release-notes/3.0-scram/#upgrade-mongodb-cr-to-scram), essentially: `use admin, db.adminCommand({authSchemaUpgrade: 1});`.  You may need to delete and re-add your oplog reader user.
+
+<h3 id="node-gyp-gotchas">node-gyp</h3>
+
+Make sure this is installed so you can compile your node modules to native if necessary.  See [Install `node-gyp`](https://github.com/nodejs/node-gyp#installation)
```
tmeasday: I don't think you should need to install `node-gyp` yourself? What happened before you did that?

content/1.4-migration.md
```diff
@@ -64,3 +64,44 @@ if (Meteor.isClient) {
...
+
+<h3 id="clear-out-oldstuff-gotchas">Clear out old stuff</h3>
+
+Try a `meteor reset` or better yet `rm -rf .meteor/local`.  Follow this with the [rebuild command](https://guide.meteor.com/1.4-migration.html#binary-packages-require-build-toolchain).
```
tmeasday: I mentioned this wrt the mongo 3.2 upgrade already, I think that's OK. Most people know to `meteor reset` in general when updating I think

content/1.4-migration.md
```diff
@@ -64,3 +64,44 @@ if (Meteor.isClient) {
...
+
+<h3 id="npm-bcrypt-gotchas">npm-bcrypt</h3>
+
+You may run into this [`npm-bcrypt`](http://stackoverflow.com/questions/29340510/bcrypt-is-breaking-my-meteor-application-how-do-i-fix-it/29347616) problem
```
tmeasday: I think we might need to say something more subtle about `npm-bcrypt`...

content/1.4-migration.md
brucejo75: https://github.com/meteor/meteor/issues/7437, I added node-gyp on @benjamn recomendation
others are seeing the issue too. https://forums.meteor.com/t/error-while-updating-to-meteor-1-4/27007/25

content/1.4-migration.md
```diff
@@ -64,3 +64,44 @@ if (Meteor.isClient) {
...
+
+<h3 id="debugger-gotcha">debugger statement issue</h3>
+If you have `debugger` statements in your code they will now hit the breakpoint even without a debugger attached.
```
tmeasday: I think we could probably just have a quick one liner about this, I don't think it's that common for people to leave debugger statements in their code.

content/1.4-migration.md
brucejo75: Your comment wrt to mongo 3.2 was regarding the meteor development DB.  I have set the MONGO urls and am using a shared DB in my development running as a windows service.  I still needed to do a reset.

I have gotten into the habit of doing a meteor reset.  But I never knew/saw it as part of the instructions.  Always worth a reminder I think?

content/1.4-migration.md
brucejo75: oh :-).

The main point was that I needed to delete it from the packages and reinstall with the new `node-gyp` I had installed.

tmeasday: Yeah, I'll chase this one up for a definitive answer.

content/1.4-migration.md
tmeasday: I mean you shouldn't *need* to do it. Do you know why you had to do it?

content/1.4-migration.md
brucejo75: agree, but the subtle change to the way the debugger statements worked is fundamental to my debug strategies.  I like the new way it works and I think it will save time.  I am ok with the more detailed info in the forum.

content/1.4-migration.md
brucejo75: hmmm, not very scientific for you.  I got a strange error.  reset/rm -rf and got a new strange error :-).  so it may have resolved something.  Always good hygiene?

content/1.4-migration.md
tmeasday: Yeah, it should be part of 1.4

tmeasday:
Thanks for this! I picked out some of it in 68fa3cdc19b739d2fb81620ecf964ea294d038d4

content/1.4-migration.md
brucejo75: hmmm, the [instructions ](https://github.com/nodejs/node-gyp#installation)are that this is a global install of node-gyp.  I did not think that meteor installed global packages.

I keep seeing threads where people are discovering that installing this can sort out their issues.  I think it should be in there...

brucejo75: Hi @tmeasday, I believe I have been confused about meteor interaction with npm for some time.  If you could look at https://forums.meteor.com/t/meteor-and-the-npm-global-repository/27154 and make a comment, I would like to know if I am thinking correctly about it now.

node-gyp is in the meteor global npm repository.

Thanks!

Issue #527, opened 2016-07-27
Provide correct Node versions for different Meteor versions


abernix:
Things have changed, need to update a couple things regarding Node versions!

(Incoming PR)

tmeasday:
Feel free to just send PRs without issues to the documentation repos @abernix

PR #528, opened 2016-07-27
Clarify Custom Deployment Node Versions

abernix:
This adds:

 * Proper versions of Node required for different Meteor versions.
 * Improved instruction for `MONGO_URL` with link to Mongo docs to explain where it is.  This is the _only_ place where `MONGO_URL` is explained.

 Closes meteor/guide#527

### DIFF
--- a/content/deployment.md
+++ b/content/deployment.md
@@ -170,15 +170,27 @@ npm install --production
 meteor build /path/to/build --architecture os.linux.x86_64
 ```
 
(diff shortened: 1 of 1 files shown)


tmeasday:
`MONGO_URL` is now explained here though: http://docs.meteor.com/api/collections.html#mongo_url

I think maybe we should link there?

abernix:
I stand corrected; I'd missed that commit!  Should we cross link them?

Issue #529, opened 2016-07-27
Mup support has changed with Meteor 1.4


abernix:
Need to update `mup` deployment options for Meteor 1.4

(Incoming PR)

PR #530, opened 2016-07-27
Provide correct mup options for various Meteor versions

abernix:
Meteor 1.4 does not work with `mupx` without forking it and providing your own docker image.  A better option is to use the kadirahq/mup developement version (also named `mup`).  A dockerImage is still necessary but hopefully the project will provide an option.  A work around is available in meteor/meteor#7475 but I'd like to avoid suggesting it publically at the moment.

Closes meteor/guide#529


tmeasday:
Thanks @abernix. PS I gave you commit rights to the guide and docs repos! Thanks so much for all your contributiosn.

PR #532, opened 2016-07-29
Fix link to sinonjs.org

vladgolubev:


abernix:
Thanks @vladgolubev! :smile:

PR #534, opened 2016-07-31
Fix broken documentation links

lvnr:
Replaced 2 links pointing to old Meteor documentation: 

One was being redirected to the new documentation automatically, but the seconds one was just displaying the first page of the documentation.


tmeasday:
Thanks @lvnr

PR #536, opened 2016-08-02
Typo fix

tsdorsey:
TODO:

"If see a failed..." should be "If you see a failed..."


lorensr:
Thanks 😊

PR #539, opened 2016-08-06
Add a note about the `replicaSet` argument

lvnr:
In some cases the `replicaSet` argument should be omitted in MongoDB connection strings.


tmeasday:
Thankyou @lvnr

PR #541, opened 2016-08-08
Naming g++ as necessary C compiler on Linux

fknipp:
I had already gcc installed, but I had to install g++ to make it working. So, I suggest naming this compiler could help others.

### DIFF
--- a/content/1.4-migration.md
+++ b/content/1.4-migration.md
@@ -25,7 +25,7 @@ In order for this rebuilding to work, you will need to install a basic compiler 
 
  - Windows users should install the [MS Build Tools](https://www.microsoft.com/en-us/download/details.aspx?id=48159).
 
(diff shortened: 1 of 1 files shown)


abernix:
Hmm, I know it's `g++` on Ubuntu, but is this true for all Linux flavors?

fknipp:
It is g++, if it is the C++ compiler of GNU.

node-gyp says **C/C++ compiler toolchain** (https://www.npmjs.com/package/node-gyp) as prerequisite, maybe this wording would be better than just **C compiler**.

abernix:
I don't know what the perfect wording would be but I do believe that you're right that `g++` is ideal to say here.  I think it's possible to get away with `gcc` in some cases, but might depend on the NPM which is being compiled.  Thanks for the PR.

PR #542, opened 2016-08-12
Update CDN web font handling instructions

brianlukoff:
Add woff2 to the list of supported extensions for webfonts.


tmeasday:
Thanks!

PR #543, opened 2016-08-12
fix links to build plugin documentation

maxnowack:
There was an unnecessary redirect to github and then back to the docs


tmeasday:
Thanks!

PR #550, opened 2016-08-22
Show how to import ValidatedMethod

emgee3:
None of the examples show how to import the validated method package so added this line to the first example.


lorensr:
Good point, thank you!

lorensr:
Wrong button :)

PR #557, opened 2016-09-07
Dead link to guide.meteor.com/dont-use-profile

jbergenthal-pwt:
Should be https://guide.meteor.com/accounts.html#dont-use-profile instead


tmeasday:
Thanks!

PR #559, opened 2016-09-15
note added to testing section

TimeBandit:
TODO:

This had me stumped while using practical-meteor:mocha until I realised what was going on. I do not see this restriction when running mocha tests with an npm package I have authored, so in that case it might be a bug as I don’t believe that the file name should be so tightly coupled to the test suite name. Please advise.

### DIFF
--- a/content/testing.md
+++ b/content/testing.md
@@ -102,7 +102,7 @@ describe('my module', function () {
 })
 ```
 
-Note that arrow function use with Mocha [is discouraged](http://mochajs.org/#arrow-functions).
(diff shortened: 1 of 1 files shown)


TimeBandit:
I have created a PR against [practicalmeteor:mocha](https://github.com/practicalmeteor/meteor-mocha/issues/69) so I think you can make this low priority for now as I do not think anyone else has raised it against the docs.

lorensr:
I saw the issue you linked is closed – should this be closed as well?

abernix:
Agree.  The change isn't true anyhow.

PR #564, opened 2016-09-22
Include password reset token expiration in breaking changes

alanning:
TODO:


lorensr:
Thanks! 🙌

Oh hey, that was for an issue I opened! 😄

alanning:
Dude!  I know!   heh, was just laughing about that with my partners yesterday.  Caused us sooooo much trouble.  And I think you ran across it as part of the work on roles-restricted which we sponsored...

lorensr:
Oops, sorry 😄😜

alanning:
Nah, just ironic.  Was a good suggestion and change.

Issue #561, opened 2016-09-17
There is no pathFor template helper in flow-router by default?


lehtu:
Hi,

I just found problem in the documentation: [useraccounts-customizing-routes](https://guide.meteor.com/accounts.html#useraccounts-customizing-routes)

There is this section:
```
<div class="btns-group">
  <a href="{{pathFor 'signin'}}" class="btn-secondary">Sign In</a>
  <a href="{{pathFor 'join'}}" class="btn-secondary">Join</a>
</div>
```
What I found is that you need to have something like [arillo:flow-router-helpers](https://atmospherejs.com/arillo/flow-router-helpers) installed to get pathFor template helper.

Cheers!

abernix:
Thanks for reporting this!  I opened PR #570 to hopefully clear this up.  Can you take a look and see if that change would have made this more clear for you?

PR #570, opened 2016-10-11
Explain that the `pathFor` helper is provided by a package

abernix:
Previously, the guide suggested using the `pathFor` helper, but didn't mention where it might have come from.  This clarifies that the `arillo:flow-router-helpers` package uses it.

Also, moved one existing paragraph up above this since I believe it created a more clear separation.

Closes meteor/guide#561


tmeasday:
Great stuff @abernix

PR #572, opened 2016-10-12
Remove note about peer dependencies testing limitation

abernix:
As found in meteor/meteor#7500, it is no longer true that you need to have a separate scaffolding test application to accomplish this.


tmeasday:
Oh, awesome!

PR #577, opened 2016-11-04
Explicitly mention contribution requirements

sebakerckhof:
Apparently, contributing to Meteor (https://github.com/meteor/meteor/pull/7858#issuecomment-257818452) requires you to follow the airbnb style guide, which was not really clear.

### DIFF
--- a/content/code-style.md
+++ b/content/code-style.md
@@ -73,7 +73,7 @@ All of the code samples in this guide and future Meteor tutorials will use all o
 
 <h3 id="style-guide">Follow a JavaScript style guide</h3>
 
-We recommend choosing and sticking to a JavaScript style guide and enforcing it with tools. A popular option that we recommend is the [Airbnb style guide](https://github.com/airbnb/javascript) with the ES6 extensions (and optionally React extensions).
+We recommend choosing and sticking to a JavaScript style guide and enforcing it with tools. A popular option that we recommend is the [Airbnb style guide](https://github.com/airbnb/javascript) with the ES6 extensions (and optionally React extensions). It's the style guide used by MDG and the one you're required to use when contributing to Meteor.
 
 <h2 id="eslint">Check your code with ESLint</h2>
 


lorensr:
Right place, since Contributing.md links here https://github.com/meteor/meteor/blob/devel/Contributing.md#submitting-pull-requests

Wording - what about "It's the style guide used for Meteor itself."? or Meteor internals, or ...

lorensr:
Let's change Contributing.md instead of linking here. @benjamn should we ask contributors to use Airbnb style guide with ES6 extensions? cc @abernix

abernix:
I'd agree.  Although we might have to be careful with saying ES6 as there still parts of the tool which cannot use ES6.  It seems that linking contributing to here was a temporary measure to deal with the discontinuation of the Meteor Wiki.  Instead of adding the extra wording here in the guide (which is intended to refer to apps), it might be best to just [directly state here in Contributing.md](https://github.com/meteor/meteor/blame/7bd2c57b1dd35f92284c59cc9300f65c3c0cbf4a/Contributing.md#L233), something along the lines of:

* New contributions should follow the [Airbnb style guide](https://github.com/airbnb/javascript) as closely as possible.
* Do not change/update code that doesn't directly relate to the feature/bug that you're working on.
* Use a short and helpful commit title (max 80 characters)
* Use a commit description which clearly explains the change and reference related issues by number (e.g. `Fixes #9999`)

abernix:
[`Contributing.md`](https://github.com/meteor/meteor/blob/devel/Contributing.md#submitting-pull-requests) no longer points here thanks to https://github.com/meteor/meteor/commit/36b4fae9e51eea2a25fb25ab26742940ffe1bfd7 and because of the existence of the new [`Development.md`](https://github.com/meteor/meteor/blob/devel/Development.md) (PR https://github.com/meteor/meteor/pull/8267).

I think it's safe to close this now without merging.  Thank you for suggesting this in the first place, @sebakerckhof.  Issues such as this were certainly inspiration for me to work on `Development.md`.  If you have any further suggestions/ideas (or objections to closing this), do tell! 😄 (Though, maybe they should go into the new document!)

PR #580, opened 2016-11-17
Small capitalisation inconsistency

alprestedge:
Small inconsistency with the Mongo collection naming conventions as can be seen here: https://docs.mongodb.com/v3.2/reference/method/db.createCollection/

And discussed here:
https://forums.meteor.com/t/collections-and-schemas/19660/37

The line as it currently stands would create a capitalised collection "Todos" instead of the lowercase "todos" as the Mongo documentation suggests.

I think it makes sense to maintain consistency between the Meteor and MongoDB documentations

### DIFF
--- a/content/collections.md
+++ b/content/collections.md
@@ -28,7 +28,7 @@ In this article, we'll look closely at how collections work in various places in
 When you create a collection on the server:
 
 ```js
-Todos = new Mongo.Collection('Todos');
+Todos = new Mongo.Collection('todos');
 ```
 
 You are creating a collection within MongoDB, and an interface to that collection to be used on the server. It's a fairly straightforward layer on top of the underlying Node MongoDB driver, but with a synchronous API:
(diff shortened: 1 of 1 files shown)


lorensr:
Looks like Sashko [prefers matching JS var](https://forums.meteor.com/t/collections-and-schemas/19660/46?u=loren). 

My inclination is consistency w/ the database docs is more important than matching JS var. When I'm developing queries, I'm looking at the Mongo docs a lot. It would be weird looking at MongoChef and seeing capitalized tables. Only table I remember seeing capitalized in my consulting is MeteorToys 😊 

Anyone else want to weigh in?

abernix:
If I needed to make a vote on it, it would be for a camelCased collection name.  That is based on the fact that the Mongo docs do it exclusively and many completely non-authoritative documents that I've read do as well:

    Todos = new Mongo.Collection('todos');

While this will likely be the only reference within Meteor itself, this `camelCase` seems more natural when when encountering the Mongo console (which is JavaScript) and typically the way the naming conventions would be for JS too:

    > db.todos.find({ ... })
    > db.someOtherCollection.find({ ... })

I guess an argument could be made for capitalized names being a way to differentiate between actual `db` methods (i.e. `db.getSiblingDB()`), but for some reason Mongo hasn't adopted that.

lorensr:
Thanks @alprestedge ! 😊

PR #582, opened 2016-11-25
dead link react-components.com so updating

phocks:
TODO:

Updated to js.coach because it was the first link I found that seemed to be as described.


lorensr:
Thank you! 😄

PR #584, opened 2016-12-01
Suggest better search for npm packages - npms.io

dandv:
https://npms.io/about


lorensr:
Looks great, thanks 😊

PR #592, opened 2016-12-22
Changed dalgard:jade into pacreach:jade

juho:
dalgard:jade is out of date and doesn't work with imports and local packages. We made a fork off it with the upstream master changes included at pacreach:jade.

### DIFF
--- a/content/build-tool.md
+++ b/content/build-tool.md
@@ -58,7 +58,7 @@ The aptly named `blaze-html-templates` package that comes with every new Meteor 
 
 <h3 id="blaze-jade">Blaze Jade templates</h3>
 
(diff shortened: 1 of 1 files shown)


lorensr:
@dalgard thoughts?

juho:
I'm happy to merge the changes back into dalgard, doesn't seem like he's been around though

juho:
@lorensr I think we can take 20 days from notification as him being AFK. No activity on GH since September either. Hope he's ok! Would probably be easier to deal with switching it back if/when he reactivates than waiting with the broken package link in the guide. :)

lorensr:
Thanks!

PR #593, opened 2016-12-29
Correct the 4th condition for the pagination state.

szchenghuang:
Add an extra rule `requested === count` for the 4th state to differentiate from the 2nd state.


lorensr:
Thanks! ☺️

PR #599, opened 2017-01-06
Added a list of environment variables

juho:
Added a list of environment variables under the "Environment variables and settings" section.

### DIFF
--- a/content/deployment.md
+++ b/content/deployment.md
@@ -50,6 +50,40 @@ Environment variables are used to set process specific things, which could conce
 
 A final note on storing these settings: It's not a good idea to store settings the same repository where you keep your app code. Read about good places to put your settings in the [Security article](security.html#api-keys).
 
+<h4 id="environment-variable-list">List of environment variables</h4>
+
+Here's a list of the environment variables you can provide to your application.
+
+`BIND_IP` 
+Bind the application server to a specific interface, for example: `BIND_IP=192.168.0.2`.
+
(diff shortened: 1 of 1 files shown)


content/deployment.md
```diff
@@ -50,6 +50,40 @@ Environment variables are used to set process specific things, which could conce
...
+`BIND_IP` 
+Bind the application server to a specific interface, for example: `BIND_IP=192.168.0.2`.
+
+`DISABLE_WEBSOCKETS` 
+Set this to 1 to disable WebSocket usage in your application.
+
+`HTTP_FORWARDED_COUNT` 
+Set this to however many network hops you have running before your Meteor application. For example, if have a nginx server acting as a proxy before your Meteor application, you would set `HTTP_FORWARDED_COUNT=1`. If you have a load balancer in front of that nginx server, the count is 2.
+
+`MAIL_URL`
```
abernix: Some of these, like `MAIL_URL` are already defined in other places, although I think this might be a better description!  Regardless, it might make sense to not double-up the descriptions (extra work to update in the future), so maybe some of those should link here?

juho:
Yeah, you might be right that the correct place for the list is in the docs and there should be a link to that here instead. I think it would fit under the command line section as its own page. Whaddya think?

juho:
ping @abernix  :) should I move this one over to docs?

abernix:
@juho Sorry for the delay!  Yes, I think moving it over to an "Environment variables" section under Command Line would be great!

PR #601, opened 2017-01-13
Fix advanced auth example in data loading.

MasterAM:
Currently the example code does not work and if it did, it would leak
sensitive info to the client.

This commit fixes those issues and adds a note regarding the security
and performance implications of using this snippet.

To point out the issues in a more detailed fashion:
- The queried field should be `_id` rather than `userId`.
- Unless the fields fetched are limited, there are 2 implications:
  1. The entire user document is published to the client (which is a security issue).
  1. The computation will be re-run every time any change is made to the user document.
- The children's `find` methods should return a cursor and not a list of cursors. Therefore, there should be 2 children and each should have the appropriate `find()` method.

I also specified the `user` parameter in the `find` functions to hint that the results of parent computations are passed to the children.

TODO:

### DIFF
--- a/content/data-loading.md
+++ b/content/data-loading.md
@@ -461,20 +461,26 @@ Meteor.publishComposite('Todos.admin.inList', function(listId) {
   const userId = this.userId;
   return {
     find() {
-      return Meteor.users.find({userId, admin: true});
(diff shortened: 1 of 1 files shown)


lorensr:
Looks great, thanks!

Ref: https://github.com/englue/meteor-publish-composite#meteorpublishcompositename-options

PR #606, opened 2017-01-20
Add a comma for readability

michielbdejong:
This old system still works, so to opt-in to the new module system
->    ,
 code must be placed inside ...


lorensr:
Thanks 😊

PR #611, opened 2017-02-09
Rename parameters of createContainer to props in the sample code.

szchenghuang:
This becomes intuitive, and consistent with the example given for Flow-Router  in the Routing section.

### DIFF
--- a/content/react.md
+++ b/content/react.md
@@ -215,8 +215,8 @@ import { Lists } from '../../api/lists/lists.js';
 import { createContainer } from 'meteor/react-meteor-data';
 import ListPage from '../pages/ListPage.js';
 
-export default ListPageContainer = createContainer(({ params }) => {
-  const { id } = params;
+export default ListPageContainer = createContainer(props => {
+  const { id } = props;
   const todosHandle = Meteor.subscribe('todos.inList', id);
   const loading = !todosHandle.ready();
   const list = Lists.findOne(id);
@@ -237,7 +237,7 @@ The container component created by `createContainer()` will reactively rerender 
 Although this `ListPageContainer` container is intended to be instantiated by the React Router (which passes in the props automatically), if we did ever want to create one manually, we would need to pass in the props to the container component (which then get passed into our data function above):
 
 ```jsx
-<ListPageContainer params={{id: '7'}}/>
+<ListPageContainer id={7}/>
 ```
 
 <h3 id="preventing-rerenders">Preventing re-renders</h3>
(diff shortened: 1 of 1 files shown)


lorensr:
this syntax looks nicer to me, @hwillson @abernix either of you more familiar w/ react?

hwillson:
Hmm - since the number of `<ListPageContainer />` props is known (and small), in this case I would usually take things one step further and handle the destructuring as the function parameters come in. So instead of:

```js
export default ListPageContainer = createContainer(props => {
  const { id } = props;
```

I would use:

```js
export default ListPageContainer = createContainer(({ id }) => {
  // now id is good to go
```

Saves an extra line, but either approach works well (and both are more clear than the current approach in the Guide and in `todos`). Speaking of which, if this is changed here it will also have to be updated in the [`todos` react branch](https://github.com/meteor/todos/tree/react). PR's welcome! 🙂

szchenghuang:
While a function is passed one, two or just very few parameters, inline destructuring does look appearing. But the introduction of `createContainer` in the guide better comes with its crystal clear usage. While the following two forms doesn't make significant difference to me,

```
export default ListPageContainer = createContainer(props => {
```

```
export default ListPageContainer = createContainer(({ id }) => {
```

, this one in the current guide is a bit confounded:
```
export default ListPageContainer = createContainer(({ params }) => {
  const { id } = params;
```

It is like being said, "some parameter in parameters is named parameters". :confounded: You get the point, and that is why this pull request was fired.

abernix:
@szchenghuang Thanks for opening this.  You're absolutely right that it's confusing as-is.  Your initial change was fine, but I would propose we take it a step further and go with @hwillson's proposal:

```js
export default ListPageContainer = createContainer(({ id }) => {
```

szchenghuang:
I just updated this request per the suggestion.

While I was about to open a pull request in the [todos react branch](https://github.com/meteor/todos/tree/react) reminded by @hwillson, I found something interesting. Consinder the [example](https://github.com/reactjs/react-router-tutorial/tree/master/lessons/06-params) given by react-router

```
<Route path="/repos/:userName/:repoName" component={Repo}/>

export default React.createClass({
  render() {
    return (
      <div>
        <h2>{this.props.params.repoName}</h2>
      </div>
    )
  }
})
```

The router collects up URL parameters and pass down to the rendering component via a single variable `params`. This makes all the sense as a router per se. The `todo react` branch is routed with `react-router` in the way described above, making it simply natural to write in its current implementation.

```
<Route path="lists/:id" component={ListPageContainer} />

export default ListPageContainer = createContainer(({ params }) => {
  const { id } = params;
```

As a result I chose not to propose it in that `react` branch.

If there is any further suggestion, please let me know.

lorensr:
Yeah, let's leave todos as is:

https://github.com/meteor/todos/blob/react/imports/ui/containers/ListPageContainer.jsx#L6

Thanks for the PR!

PR #612, opened 2017-02-15
Refactor and expand production MongoDB section

lorensr:
any comments? moved mongo to a new h2 section and added a couple paragraphs

### DIFF
--- a/content/deployment.md
+++ b/content/deployment.md
@@ -60,7 +60,7 @@ What URL will users use to access your site? You'll probably need to register a 
 
 <h3 id="ssl">SSL Certificate</h3>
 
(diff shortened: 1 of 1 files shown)


abernix:
It would appear that the notable addition here is the section on running your "Own server" for MongoDB?

If so I'm not really in favor.  The guide was intended to be an opinionated, best-practice, recommended-configuration guide.  By providing the option for "Own server" it almost implies that we're recommending that as a viable option.  I'm really not sure we should be recommending anyone to run their own MongoDB deployment or doing anything besides discouraging it.  We already mention all the advantages (bullet-points) of a hosted provided.  My interpretation of running my own server after reading that is: "Yup, I don't want to do that".

I just think this begs the question of: Why add "Own server" at all since it's clearly not recommended.

lorensr:
I think it's the right choice in some circumstances. I remember one of the bigger meteor companies like workpop or someone said there was a large cost difference between a sizable DO installation and compose, which made them switch to the former. 

I've also worked for meteor companies that default to a DO Mongo template image. I think having two subsections and describing the second a little makes it more clear that yes, we've considered their option, and we still recommend hosted. 

For app server, we recommend Galaxy but still mention the own server option for completeness.

lorensr:
This would also fix #620

PR #613, opened 2017-02-28
Added paragraph on TypeScript conditional imports.

realyze:
TODO:

### DIFF
--- a/content/build-tool.md
+++ b/content/build-tool.md
@@ -68,7 +68,7 @@ A guide on installation, configuration and usage of TypeScript with Meteor is cu
 
 TypeScript does not support nested `import` statements, therefore conditionally importing modules requires you to use the `require` statement (see [Using `require`](https://guide.meteor.com/structure.html#using-require)).
 
-To maintain type safety, you can take advantage of TypeScript's import elision and reference the types using the `typeof` keyword. See the [TypeScript handbook article](https://www.typescriptlang.org/docs/handbook/modules.html#optional-module-loading-and-other-advanced-loading-scenarios) for details.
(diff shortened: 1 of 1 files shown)


realyze:
https://github.com/meteor/meteor/issues/8428

abernix:
This looks great!  Thank you for taking the time do this (and research it!).

Did you have any additional feelings about this update after reading the information about dynamic imports in https://github.com/meteor/meteor/pull/8327#issuecomment-284160678?

If not, I'm happy to merge this!

realyze:
Thanks @abernix . The bottomline of https://github.com/meteor/meteor/pull/8327#issuecomment-284160678 is basically that adding dynamic `import` to TypeScript will be quite a challenge for the TS team so it probably won't happen anytime soon (if at all).

When Meteor 1.5 is released, I guess it will make sense to update this by mentioning the `module.dynamicImport` workaround.

realyze:
BTW I also wrote a short post about it which explains it with a Meteor code example [on my blog](http://ideasintosoftware.com/typescript-conditional-imports/) but I didn't want to link to my personal blog from the guide. Dunno if that's frowned upon or legit...you tell me, I guess. :-)

abernix:
@realyze If it's relevant and we don't have a good place to put it right now (both of these things seem true), go for it – I approve of [the link you've provided](http://ideasintosoftware.com/typescript-conditional-imports/).

I will make the disclaimer that at some point and time in the future, I could see us having a have a dedicated TypeScript section where this would fit very well (although, hopefully at some point we can make this exception obsolete). :smile:

realyze:
@abernix Cool, I've updated the patch.

Having a dedicated (and much larger, yes :) ) TS section would be awesome. It certainly appears TypeScript's adoption is increasing quite rapidly and I believe having an easy to follow and readable guide for TS would be encouraging to new Meteor users. There definitely are some gotchas when working with Meteor/TS combo.

lorensr:
Thanks @realyze ! 😊

michaeljota:
FYI: TS 2.4 add supports for Dynamic Imports :+1:

PR #616, opened 2017-03-03
Add a notice of the exclusion of *.min.js from minification.

szchenghuang:
This piece of information is really helpful when it is needed, and ought to be conveyed in the guide. The implementation can be found in this [commit](https://github.com/meteor/meteor/pull/6986).


lorensr:
Thanks! 😊

szchenghuang:
Thank you for polishing, @lorensr.

PR #614, opened 2017-02-28
Update dispatch test driver package descriptions

aldeed:
Update to reflect the merger of dispatch:mocha-phantomjs into dispatch:mocha

PR #618, opened 2017-03-05
Update the test driver packages in the testing page

vjvasilev95:
TODO:


vjvasilev95:
I just noticed that #614 has already covered the change in the listing of the test driver packages, so this part might be considered a duplicate, but the Continuous Integration setup change has not been covered there, so this PR is still valid.

PR #621, opened 2017-03-08
Improve TypeScript section

birkskyum:
Added links to more thorough external resources on how to use TypeScript with Meteor.

TODO:


lorensr:
Thank you! 😊

PR #622, opened 2017-03-09
Removing list from the returned values

MehdiAlouafi:
Removing list from the returned values because it raises this
error : '
  Warning: you are returning a Mongo cursor from getMeteorData.
  This value will not be reactive. You probably want to call `.fetch()`
  on the cursor before returning it.'

TODO:

### DIFF
--- a/content/react.md
+++ b/content/react.md
@@ -222,7 +222,6 @@ export default ListPageContainer = createContainer(({ id }) => {
   const listExists = !loading && !!list;
   return {
     loading,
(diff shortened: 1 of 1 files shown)


lorensr:
Looks like `list` should be an object, not a cursor:

`const list = Lists.findOne(id);`

Is there something else that might be causing that error?

lorensr:
Closing for now, feel free to re-open if it's still an issue.

PR #625, opened 2017-03-25
Update react.md

Guikingone:
Add the 'history' library call and the Router new way of declaring the routes.

TODO:

### DIFF
--- a/content/react.md
+++ b/content/react.md
@@ -334,6 +334,56 @@ Meteor.startup(() => {
 });
 ```
 
+Since the version 4.0.0 of react-router, you don't have any longer access to the browserHistory object from the library, to do so, you first need to install history : 
+
+```bash
+npm install --save history
+```
+
+Once it's done, you can call the object and use the same syntaxe as earlier (except for the declaration of the Route who don't accept any longer to have multiples childrens in a single instance of Router) : 
+
+```js
+import React from 'react';
+import { Router, Route } from 'react-router';
+
+// Call the library history
+import createBrowserHistory from 'history/createBrowserHistory';
+
+// route components
+import AppContainer from '../../ui/containers/AppContainer.js';
+import ListPageContainer from '../../ui/containers/ListPageContainer.js';
+import AuthPageSignIn from '../../ui/pages/AuthPageSignIn.js';
+import AuthPageJoin from '../../ui/pages/AuthPageJoin.js';
+import NotFoundPage from '../../ui/pages/NotFoundPage.js';
+
+const browserHistory = createBrowserHistory();
+
+export const renderRoutes = () => (
+  <Router history={browserHistory}>
+    <div><!-- Needed for have multiples childrens -->
+      <Route path="/" component={AppContainer}>
+        <Route path="lists/:id" component={ListPageContainer}/>
+        <Route path="signin" component={AuthPageSignIn}/>
+        <Route path="join" component={AuthPageJoin}/>
+        <Route path="*" component={NotFoundPage}/>
+      </Route>
+    <div>
+  </Router>
+);
+```
+
+The call of this constant doesn't change : 
+
+```js
+import { Meteor } from 'meteor/meteor';
+import { render } from 'react-dom';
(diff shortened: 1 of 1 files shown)


lorensr:
Thanks for the update! So `<Router>` can only have one child? Isn't the current `<Route path="/"` a single child?

Guikingone:
@lorensr In fact, it depends on the way you declare the routes, if you declare the routes using the self-closed approach : 

```js
  <Router history={browserHistory}>
      <div>
        <Route path="/" component={StartForm} />
        <Route path="/tickets" component={TicketsForm} />
        <Route path="/contact" component={Contact} />
      </div>
  </Router>
```

The Router does not accept the Single Child approach and all the routes must be enclosed in a div, if you use the old approach : 

```js
const routes = (
  <Router history={browserHistory}>
          <Route path="/" component={StartForm}>
              <Route path="/tickets" component={TicketsForm} />
              <Route path="/contact" component={Contact} />
          </Route>
  </Router>
);
```

The compiler works but the "browser history" gonna refuse to have one element who have X childrens, don't know why this change and how can we overcome this type of declaration (maybe with Typescript support ?) but the new approach seems to be "cleaner" in my opinion and way more simple to understand for new comer.

lorensr:
@Guikingone sorry for the delay in getting back to you—the PR looks like a combination of those two options. Should we make the root path self-closing? Would also be more in line with [the docs](https://reacttraining.com/react-router/web/example/basic). And we could add [exact](https://reacttraining.com/react-router/web/api/Route/exact-bool).

```
      <Route exact path="/" component={Home}/>
      <Route path="/about" component={About}/>
      <Route path="/topics" component={Topics}/>
```

Guikingone:
As I saw from now, yes, we can use 'exact' in the Route declaration, I've update the PR with the modifications.

lorensr:
Thanks. Looks like this would also be more in line with the doc example?

```
<div>
  <Route exact path="/" component={AppContainer}/>
  <Route path="lists/:id" component={ListPageContainer}/>
  <Route path="signin" component={AuthPageSignIn}/>
  <Route path="join" component={AuthPageJoin}/>
  <Route path="*" component={NotFoundPage}/>
</div>
```

Guikingone:
Exact, sorry for the bad syntax, was a hotfix.

lorensr:
Thanks again @Guikingone 😊

jeroenhoebe:
Currently I'm rewriting the authorization code for my Meteor (v1.6.1) project with React-router (v4.2.2) and I stumbled on two things. First of, related to issue #681, I don't thing the guide is correct. This indeed doesn't work: 
```
<div>
  <Route exact path="/" component={AppContainer}/>
  <Route path="lists/:id" component={ListPageContainer}/>
  <Route path="signin" component={AuthPageSignIn}/>
  <Route path="join" component={AuthPageJoin}/>
  <Route path="*" component={NotFoundPage}/>
</div>
```
Shouldn't it be like this:

```
<Switch>
  <Route exact path="/" component={AppContainer}/>
  <Route exact path="/lists/:id" component={ListPageContainer}/>
  <Route exact path="/signin" component={AuthPageSignIn}/>
  <Route exact path="/join" component={AuthPageJoin}/>
  <Route component={NotFoundPage}/>
</Switch>
```

Second problem is that in the guide I can't really find how to properly use React-router with the accounts system. It probably would be a big help for everyone if there was a clear example or maybe a next step in the `Todo App with React` explaining it.
Also you can't use `AppContainer` in the same way as in FlowRouter. Because in the above example `AppContainer` wont be like a 'layout' component. 

I found a couple of authentication examples online: [The Meteor Chef](https://themeteorchef.com/tutorials/getting-started-with-react-router-v4) and one on [StackOverflow](https://stackoverflow.com/a/48913629) but the one from StackOverflow uses onEnter and that's not React-router v4 but v3. There are more examples but they're using older version of react-router.

lorensr:
Thanks @jeroenhoebe! I've fixed the code and added this issue: #753

PR #627, opened 2017-04-07
Update Code on form validation

bastiW:
because of 
https://forums.meteor.com/t/implementing-validation-with-mdg-validate-method-display-the-validation-error/35475

I have updated the template helper

TODO:


lorensr:
Quite right, thanks for finding this 😊

PR #628, opened 2017-04-24
Added info about installing older version of Android SDK Tools

maciejsaw:
Because It's not obvious where to find older versions and how to install them.

TODO:


lorensr:
Thanks! In addition to CLA, do you know the locations of the linux and windows versions of the tools?

maciejsaw:
I think you just need to change the suffix of the URL - good point it's worth linking all the platforms or at least mentioning the suffixes
https://dl.google.com/android/repository/tools_r25.2.3-linux.zip
https://dl.google.com/android/repository/tools_r25.2.3-windows.zip

PR #630, opened 2017-05-03
Crosswalk no longer maintained

paulincai:
TODO:

### DIFF
--- a/content/mobile.md
+++ b/content/mobile.md
@@ -63,6 +63,8 @@ This threatens to leave many older Android devices behind however, because they 
 You can add the Crosswalk plugin to your app with `meteor add crosswalk`.
 > If you receive an error message trying to run the app on your device after adding or removing the Crosswalk plugin, you may have to remove the existing app from your device first.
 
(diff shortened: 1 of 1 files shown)


abernix:
Maybe we should just remove the whole Crosswalk section altogether?

paulincai:
I feel that would be the right way, however some build tips should be provided. I dumped the crosswalk and rely now on the web view alone and…disaster. I use the cheapest Android 5.1 in the market ( Lenovo A2010) and the 2017 Samsung A3. No difference. Same choppy sliding of a card from left to right. I need to re-write and eliminate some css animation or something. It is not really like “you just need Meteor” for an Android app development any more. You need some more knowledge and downgrading of UX.

skirunman:
We did some analysis of dropping Crosswalk in our app and it still gives the best cross platform support in our opinion. I would not recommend dropping it entirely from documentation, yet, but the comment about the project not being supported going forward makes sense to me.

abernix:
Thanks for the additional insight, @skirunman!

PR #631, opened 2017-05-09
Add missing = to meteor build command

skirunman:
Add `=` build command to be consistent and match output of `meteor help build` which is `--server=https://example.com:443`

TODO:


abernix:
Thanks, @skirunman!

PR #632, opened 2017-05-09
Change reference to version of Android API

skirunman:
Meteor v1.4.4.2 ships with support for Cordova Android v6.1.1, which supports Android APIs 25, 24 and earlier so this reference is out of date. Making it more generic so the user should check with Android APIs are supported for the version of Cordova they are using.

TODO:


abernix:
Thanks!

PR #635, opened 2017-05-11
Need to manually copy Android APK with Crosswalk

skirunman:
Add comment about having to manually copy the Android APK when using Crosswalk during Android build process. See https://github.com/meteor/meteor/issues/5274

TODO:


lorensr:
Thanks! Sorry for the delay

PR #637, opened 2017-05-20
mention mhagmajer:server-router

mhagmajer:
This mentions https://atmospherejs.com/mhagmajer/server-router in the routing section.

When discussing server routing solutions people often ask about ways to authenticate users given that Meteor doesn't use cookies. It seems helpful to talk about this briefly in the guide.


abernix:
@lorensr Any thoughts on this?

lorensr:
@mhagmajer What part of your package isn't covered by restivus? I think it has authenticated custom endpoints?

lorensr:
The code looks good! Only security issue is that if connection is not https, and the single-use token is observed in transit, and the observer can get their HTTP request to the server faster than the client, which is unlikely, but possible if they kept an open TCP connection 😄

mhagmajer:
@lorensr thanks for your feedback! I'm not familiar with restivus. My goal here was to create a simple package tailored for the job and with support for Flow type definitions. On the other hand, restivus seems to incline folks more towards REST, which I think is being replaced by GraphQL as the generally recommended API protocol.

How do you feel about mentioning this in the docs? We're already been able to successfully adopt this package at our company and we would love to give back to the Meteor community. :)

StorytellerCZ:
@mhagmajer I have fixed the conflict, but please give this one more look. Also I will hold on merging this until the following PR is merged and published:
https://github.com/mhagmajer/server-router/pull/12

StorytellerCZ (approved):

mhagmajer:
Hi, @StorytellerCZ! Looks good! Thanks for taking care of the merge conflict.

PR #639, opened 2017-05-27
react router's version

zhe1ka:
since react-router has version 4 this command will install the latest version of it and react-router (4) has another logic.


lorensr:
Thanks for submitting this! I think this should be fixed in #625 by an update to the code example?

PR #644, opened 2017-06-15
Fix MAIL_URL documentation

sebakerckhof:
See https://github.com/meteor/meteor/issues/8804


abernix:
Made a few adjustments.  Hope they sound okay to you.  Thanks again!

sebakerckhof:
Yep, I think it's as correct as it can get now.

PR #652, opened 2017-07-11
Modify script lint

AurelienC:
An error occurred when the *lint* script is launched.
Adds `exit 0` for a proper stop of script.

TODO:


lorensr:
From this thread I get the sense that it's not recommended for everyone, just something some people may need to add?

https://github.com/eslint/eslint/issues/2949

PR #653, opened 2017-07-12
data-loading.md: swap simple-schema for check

coagmano:
I came across this in the guide today and thought it was over complicating a basic check of function arguments. Not sure if this is a wanted change. I thought I should throw a quick PR in as a proposal to change.

The use of simple-schema is both more complicated in code as well as relying on an external dependency when the excellent check package is already part of Meteor core. The docs should be as simple as possible to make learning easier friendlier.

### DIFF
--- a/content/data-loading.md
+++ b/content/data-loading.md
@@ -77,9 +77,7 @@ Here's an example of a publication which takes a named argument. Note that it's 
 ```js
 Meteor.publish('todos.inList', function(listId) {
   // We need to check the `listId` is the type we expect
-  new SimpleSchema({
-    listId: {type: String}
-  }).validate({ listId });
+  check(listId, String);
 
   // ...
 });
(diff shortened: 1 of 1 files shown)


lorensr:
The recommendation for `simple-schema` for Methods will probably stay, but I don't think it needs to be the same for pubs. My preference is `check` in these pubs for brevity. Anyone else have thoughts?

hwillson:
I definitely like the brevity of using `check` as well, but using `SimpleSchema` here was intentional (see https://github.com/meteor/todos/issues/57). I believe the intent was to keep the argument validation approach uniform across `todos`, so `SimpleSchema` is used for publication and method argument checking, as well as collection data validation. That being said, `todos` hasn't strictly followed this approach - [`check` is still being used in places](https://github.com/meteor/todos/blob/59f6621d315a5c1527c1ef551a23a5bf7ec0e573/imports/api/todos/incompleteCountDenormalizer.js#L22).

I think it's okay to leave it as is given that the [Guide does briefly mention both approaches](https://guide.meteor.com/security.html#method-rules). That being said, if this is updated in the Guide, then it should also be updated in `todos`.

coagmano:
Ahh I missed that `todos` uses simple-schema because I've only seen the [`simple-todos` tutorial ](https://www.meteor.com/tutorials/blaze/)on the main website. Interesting that simple-todos doesn't check publication arguments at all.

What's the benefit of using simple-schema over check for simple type checking like this?
```
Meteor.publish('todos.inList', function(listId, limit) {
new SimpleSchema({
    listId: { type: String },
    limit: { type: Number }
  }).validate({ listId, limit });
```

PR #660, opened 2017-10-02
Need to ensure default profile hook still runs

ericwikman:
If you don't set the profile from the options then you lose the default hook behaviour for the profile. For instance if a user signs up via Facebook then their profile should have a name field set as their name on Facebook, but if you don't have the code I just added then there will not be a profile at all. 
http://docs.meteor.com/api/accounts-multi.html#AccountsServer-onCreateUser

TODO:


ericwikman:
did i do this correctly? first time to contribute to the docs for meteor.

lorensr:
Thanks, looks good! 😄

PR #666, opened 2017-10-28
Remove mention of react-addons-pure-render-mixin

StorytellerCZ:
Remove mentions of react-addons-pure-render-mixin which is [no longer used](https://github.com/meteor/react-packages/issues/239) from React > Using Meteor's data system > using withTracker

TODO:


lorensr:
Thanks Jan! 😊

PR #674, opened 2017-11-09
Remove refs to Kadira

merlinpatt:
I generalized "how to monitor your application" so the line doesn't need to be changed in the future. I also replaced the KADIRA_OPTIONS_HOSTNAME reference with a link to the environment variables Meteor uses.


lorensr:
Thanks! 😊

Issue #675, opened 2017-11-15
Document launch storyboard recommendation for iPad Pro


hwillson:
See https://github.com/meteor/meteor/issues/9240#issuecomment-344405768. We should consider documenting this workaround in the Guide, likely under the [App icons and launch screens](https://guide.meteor.com/mobile.html#configuring-app-icons-and-launch-screens) section.

PR #677, opened 2017-11-15
Add info on using iOS launch story board images

skirunman:
Resolves https://github.com/meteor/guide/issues/675


lorensr:
Nice, thanks!

PR #683, opened 2017-12-10
Remove outdated information

kurapikats:
Meteor.userId() can now be used both in method calls and publications since v1.5.1


lorensr:
Thanks!

PR #684, opened 2017-12-19
Corrected importing of ReactiveVar

VenomCLC:
While trying out some personal code based on this example, I found out that the correct way of importing a ReactiveVar is using curly braces around it. not without, as the docs suggest in the modified example.


hwillson:
Thanks @VenomCLC - if you can sign the [Meteor Contributor Agreement](https://contribute.meteor.com/), we'll get this merged.

abernix:
I've made my own contribution to correct this mistake. :smile:

PR #689, opened 2018-01-22
Update mobile.md

skirunman:
Add info on how to use mobile-config.js to add iOS privacy usage notification prompts, which are required to submit an app that accesses certain iOS features post iOS 10.


lorensr:
Great, thanks! 😊

PR #706, opened 2018-02-14
Use theme from npm, rather than submodule.

abernix:
The full details of this PR can be found in https://github.com/meteor/docs/pull/187.

Also, Ref: meteor/hexo-theme-meteor#51


abernix:
cc @lorensr @hwillson done and also back-ported to all the version branches.

Issue #338, opened 2016-04-03
Question: where to import secret code from method


bitsuns:
In the [Secret Code](https://github.com/meteor/guide/blob/master/content/security.md#secret-code) example on the Security section, how/where would you import the secret code (i.e., MMR)?

If this were imported in the methods module, which is loaded on the client, the import would fail.

stubailo:
Ah, that's a really good point. Unfortunately, ES2015 doesn't allow conditional imports, so we have to work around the standard a bit:

```js
let MMR;
if (Meteor.isServer) {
  MMR = require('/my/mmr/module');
}

// .. now we can use it on the server, but not on the client
```

This would be a really good thing to add both in the app structure, and methods articles. If anyone has time to put together a PR, that would be awesome.

bitsuns:
Ahh, okay. Thank you for the clarification.

I may put a PR together that includes this and some of the material on async/sync w/in methods. Granted, the async section of [using-packages](https://github.com/meteor/guide/blob/master/content/using-packages.md) covers most of what I'd like to see for the latter.

stubailo:
I'd suggest making those separate PRs just to keep the discussions separate!

stubailo:
Similar discussion here: https://github.com/meteor/validated-method/issues/40

TimeBandit:
This had me stumped to. I don't think the guide is explicit enough here w.r.t finally importing and running the secret sauce. Based on you response above I used require inside the !isSimulation block of a method definition. A few more worked on the guide would have saved me a lot of head scratching.
```

run({ amount, currency, source, metadata }) {

        if (this.isSimulation) {
            console.log('client side method simulation');
        } else {
            // only run server-side
            let charge = require('./server/charge').default();
            let result = charge({ amount, currency, source, metadata });
            return result
        }
    },
```

xvendo:
This is still not clear in the meteor guide! Please give an example of the module export and the import of MMR.

lorensr:
Does this work for you? https://github.com/meteor/guide/pull/744

xvendo:
Thank you @lorensr. Yep that's the way to go!

PR #744, opened 2018-02-28
Add import/export statements to Secret server code

lorensr:
For #338

### DIFF
--- a/content/security.md
+++ b/content/security.md
@@ -338,16 +338,23 @@ Secret business logic in your app should be located in code that is only loaded 
 If you have a Meteor Method in your app that has secret business logic, you might want to split the Method into two functions - the optimistic UI part that will run on the client, and the secret part that runs on the server. Most of the time, putting the entire Method on the server doesn't result in the best user experience. Let's look at an example, where you have a secret algorithm for calculating someone's MMR (ranking) in a game:
 
 ```js
-// In a server-only file
+// In a server-only file, for example /imports/server/mmr.js
 MMR = {
   updateWithSecretAlgorithm(userId) {
     // your secret code here
   }
 }
+
+export default MMR;
 ```
 
 ```js
 // In a file loaded on client and server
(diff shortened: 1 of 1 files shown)


content/security.md
```diff
@@ -338,16 +338,23 @@ Secret business logic in your app should be located in code that is only loaded
...
 ```
 
 ```js
 // In a file loaded on client and server
+let MMR;
+if (Meteor.isServer) {
+  MMR = require('/imports/server/mmr.js');
+}
+
 const Meteor.users.methods.updateMMR = new ValidatedMethod({
```
abernix: How does this even work?

content/security.md
```diff
@@ -338,16 +338,23 @@ Secret business logic in your app should be located in code that is only loaded
...
 }
+
+export default MMR;
 ```
 
 ```js
 // In a file loaded on client and server
+let MMR;
+if (Meteor.isServer) {
+  MMR = require('/imports/server/mmr.js');
```
abernix: Would this be an opportunity to use a dynamic `import()`?

Or alternatively, could this just be in the `} else {` below?

```js
require("/imports/server/mmr.js").updateWithSecretAlgorithm(this.userId);
```

content/security.md
lorensr: Copied from sashko. Tried, and it doesn't 😄

lorensr:
Fixed, thanks! 

Tested here: https://github.com/lorensr/demo-secret-method-code

content/security.md
abernix: Haha, I was confused by the entire thing... including this which I think is still a problem, even if not directly related to this PRs fix:

```
const Meteor.users.methods.updateMMR = new ValidatedMethod({
```

abernix: I mean, I'm positive that's a syntax error, specifically because of the `const` declaration.

PR #750, opened 2018-03-03
Update Guide to make Helmet the official recommendation for HTTP Headers

toinevk:
This PR updates the Guide to make Helmet the official recommendation for HTTP Headers, in addition the guide provides recommend config for two headers. 

TODOs:
- [x] Update [CHANGELOG.md](https://github.com/meteor/guide/blob/master/CHANGELOG.md)


hwillson:
Thanks @toinevk! I haven't had a chance to review this yet, but will shortly. 

Reference issue: https://github.com/meteor/meteor/issues/9689

content/security.md
```diff
@@ -453,6 +454,84 @@ It's best to handle the redirection from HTTP to HTTPS on the platform which han
...
+WebApp.connectHandlers.use(
+  helmet.contentSecurityPolicy({
+    directives: {
+      defaultSrc: ["'self'"],
+      scriptSrc: ["'self'", "'unsafe-inline'"],
+      connectSrc: ["*"],
+      imgSrc: ["'self'"],
+      styleSrc: ["'self'", "'unsafe-inline'"],
+    },
+    browserSniff: false
```
abernix: @dhrubins If you have a moment, do you think you could share your (likely more experienced) CSP eyes 👀 with this ruleset recommendation?

dhrubins: Yeah, I agree that the default policy there is a reasonable baseline, but unfortunately CSP can get complex very quickly, so perhaps there could also be some links to CSP tools to help folks through the tough parts. Here are a few that helped me: [Google's CSP Evaluator](https://csp-evaluator.withgoogle.com/), [Report-URI's CSP Builder](https://report-uri.com/home/generate), [CSP documentation from Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src), [CSPValidator.org](https://cspvalidator.org). Plus, the awesome [SecurityHeaders.com](https://securityheaders.com/) since helmet also helps you set other headers aside from CSP.

One Meteor specific challenge we had @LegalRobot was with removing unsafe-inline, which seems like a reasonable first step from the baseline in this PR. We switched to a mixture of script hashes and origins in our script-src directive, but we quickly found that this causes the Meteor runtime config script to get blocked (I don't remember if it was the lack of unsafe-inline, or the CSP rule that once you use a hash that everything needs a hash). Anyway, it effectively blocks Meteor from loading anything to the client after that. The only way we've figured out around this is to take the hash of the runtime config upon startup (and we needed it in base64 so we used node's native crypto module rather than the Meteor sha package, as is standard in Meteor core):

```
import crypto from 'crypto'
import { Autoupdate } from 'meteor/autoupdate'

Meteor.startup(() => {
  const runtimeConfig = Object.assign(__meteor_runtime_config__, Autoupdate)
  runtimeConfig.accountsConfigCalled = true
  const runtimeConfigHash = crypto.createHash('sha256').update(`__meteor_runtime_config__ = JSON.parse(decodeURIComponent("${encodeURIComponent(JSON.stringify(runtimeConfig))}"))`).digest('base64')
  ...
})
```

Now our helmet config for the script-src directive includes 'sha256-${runtimeConfigHash}', which just feels kinda hack-y (but, hey, it works!). We also found that we had to run this in the startup callback, rather than during the build.

We also had to conditionally disable some directives when running locally. I'm documenting our whole helmet config line by line and will post it shortly to hopefully save some other folks some pain when implementing the same.

dhrubins: Here's a gist with our settings and how we use helmet: https://gist.github.com/dhrubins/31248506873a006ba5e6b459a81855c2

Writeup coming soon.

toinevk:
@abernix Are you happy to merge? We can then move to deprecate the browser policy packages.

Issue #691, opened 2018-02-01
Swap practicalmeteor:mocha for meteortesting:mocha


coagmano:
1.6.1 release notes recommend swapping `practicalmeteor:mocha` for `meteortesting:mocha`, but the guide has many references to `practicalmeteor:mocha` throughout

Not sure how much work is involved in switching, but the guide shouldn't refer to a package with known issues on 1.6.1

abernix:
I think they are almost entirely interchangeable.  Could you report back with your findings and submit a PR to make the swap if you find this to be true?

toinevk:
I have actioned this and its ready for review.

filipenevola:
I'm closing this issue because it's too old.

We are going to make many updates to the guide in the next weeks and it's better to focus on issues with recent activity. If you think this issue is still relevant please open a new one.

PR #751, opened 2018-03-03
Replace practicalmeteor:mocha with meteortesting:mocha

toinevk:
Per https://github.com/meteor/guide/issues/691, replacing practicalmeteor:mocha with meteortesting:mocha. Given the similarities this was a simple find and replace.

### DIFF
--- a/content/testing.md
+++ b/content/testing.md
@@ -375,14 +375,12 @@ In this case, the `TodoItem` component calls a [Meteor Method](/methods.html) `s
 To run the tests that our app defines, we run our app in [test mode](#test-modes):
 
 ```txt
-meteor test --driver-package meteortesting:mocha
+TEST_WATCH=1 meteor test --driver-package meteortesting:mocha
 ```
 
 As we've defined a test file (`imports/todos/todos.tests.js`), what this means is that the file above will be eagerly loaded, adding the `'builds correctly from factory'` test to the Mocha registry.
 
-To run the tests, visit http://localhost:3000 in your browser. This kicks off `meteortesting:mocha`, which runs your tests both in the browser and on the server. It displays the test results in the browser in a Mocha test reporter:
-
-<img src="images/mocha-test-results.png">
+To run the tests, visit http://localhost:3000 in your browser. This kicks off `meteortesting:mocha`, which runs your tests both in the browser and on the server. It will display the test results in a div with ID mocha.
 
 Usually, while developing an application, it makes sense to run `meteor test` on a second port (say `3100`), while also running your main application in a separate process:
 
--- a/content/testing.md
+++ b/content/testing.md
@@ -478,8 +478,8 @@ describe('Lists_show_page', function () {
   beforeEach(function () {
     StubCollections.stub([Todos, Lists]);
     Template.registerHelper('_', key => key);
-    sinon.stub(FlowRouter, 'getParam', () => listId);
-    sinon.stub(Meteor, 'subscribe', () => ({
+    sinon.stub(FlowRouter, 'getParam').returns(listId);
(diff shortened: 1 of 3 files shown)


abernix:
Anyone watching have any objections to this?

serut:
Yeap ! you should not let practicalmeteor:sinon and practicalmeteor:chai in the guide if you remove practicalmeteor:mocha. Moreover, this is wrong : "It displays the test results in the browser in a Mocha test reporter:" as meteortesting only display clients tests in the report.

abernix:
@serut:

> Moreover, this is wrong : "It displays the test results in the browser in a Mocha test reporter:" as meteortesting only display clients tests in the report.

What you're saying is true (that the browser will only show client tests and the server test will be shown in the console), but this appears to have been wrong in the guide both before and after this PR, correct?

> you should not let practicalmeteor:sinon and practicalmeteor:chai in the guide if you remove practicalmeteor:mocha.

Can you explain why you're making that mandate?  I ask since both `practicalmeteor:sinon` and `practicalmeteor:chai` are just older implementations of the npm versions of the `sinon` and `chai` testing libraries and should have nothing to do with the actual test running (which `practicalmeteor:mocha` _does_).  I agree that they should be updated (they are unnecessary).

I agree with your suggestions, but it seems like both of them could be done in a separate PR (which would be appreciated if you have time!)

serut:
> What you're saying is true (that the browser will only show client tests and the server test will be shown in the console), but this appears to have been wrong in the guide both before and after this PR, correct?

No, practicalmeteor:mocha was displaying both server and client tests. The screenshot [available here](https://guide.meteor.com/testing.html#driver-packages) is a real one (in watch mode):

Meteor guide report

This PR should bring an updated version of the screenshot, as these 2 tools are not isofunctional.

> Can you explain why you're making that mandate?

It looks like `practicalmeteor:chai` and `practicalmeteor:sinon` are looking for the same version of `coffeescript` that `practicalmeteor:mocha`. The main goal of removing `practicalmeteor:mocha` is to remove dependency to `coffeescript@1`, so we can't promote any `practicalmeteor:*` package anymore.  

Here is the error message if you integrate another module that requires `coffeescript 2` : 
```
Conflict: Constraint coffeescript@1.0.3 is not satisfied by coffeescript 2.0.1_1.
Constraints on package "coffeescript":
* coffeescript@=2.0.1_1 <- top level
* coffeescript@1.0.4 <- arillo:flow-router-helpers 0.5.2
* coffeescript@1.0.4 <- zimme:active-route 2.3.2
* coffeescript@1.0.4 <- tap:i18n 1.8.2
* coffeescript@1.0.4 <- softwarerero:accounts-t9n 1.3.9
* coffeescript@1.0.3 <- practicalmeteor:chai 2.1.0_1
* coffeescript@1.0.3 <- practicalmeteor:sinon 1.14.1_2
```
[source](https://github.com/meteor/meteor/pull/9107#issuecomment-338285855)

Of course, you can accept this PR and iterate again and again, that's just my recommendation.

serut:
@toinevk 
Don't forget to upgrade the sinon syntax, as you are upgrading sinon V1.14 to V5.0 : 
```
    sinon.stub(FlowRouter, 'getParam', () => listId);
```
becomes
```
    sinon.stub(FlowRouter, 'getParam').returns(listId);
```
Hopefully, there is only 1 or 2 occurences of sinon in that file.

And I had no change while upgrading chai, so after that LSTM :balloon: !

toinevk:
Thanks @serut I have tested and is passing. Should be ready to merge.

serut:
@abernix feel free to merge that PR.

abernix:
Thanks to all, especially @toinevk for opening this originally!

Issue #800, opened 2018-04-26
Update all react router (and associated) code samples


hwillson:
PR https://github.com/meteor/todos/pull/261 updates the `react` branch of `todos` to use `react-router` 4 (and adjusts the codebase to follow the `react-router` 4 way of thinking). Once that PR is merged, we'll need to update the `react-router` code samples in the Guide to match `todos`.

filipenevola:
I'm closing this issue because it's too old.

We are going to make many updates to the guide in the next weeks and it's better to focus on issues with recent activity. If you think this issue is still relevant please open a new one.

PR #801, opened 2018-04-27
Updates for `react-router` v4 changes

hwillson:
Fixes https://github.com/meteor/guide/issues/800.


filipenevola:
The CLA is not signed and also we don't need to explain React Router on Meteor guide as React Router works in the same way for Meteor users as for any other users.

PR #835, opened 2018-05-07
Replace Stylus Package Recommendation

toinevk:
The Stylus package has been deprecated. This change makes coagmano stylus package the recommended community package.


lorensr:
Thanks! ☺️

coagmano:
😅 I guess I have to keep maintaining it then

lorensr:
😄 didn't find any other maintained stylus packages. We can also just
deprecate stylus as a whole at some point

PR #855, opened 2018-06-05
Add 1.7 option to the docs, as well as 1.7 migration section

hwillson:
Adds the 1.7 doc dropdown and 1.7 migration section.

@benjamn @abernix Let me know if there's anything else important we should be calling out in the migration section. Thanks!


benjamn:
Can we say something about `meteor.mainModule`? Or is that too much of a distraction? Could just borrow content from `History.md`.

benjamn:
Deployed; thanks @hwillson! https://guide.meteor.com/1.7-migration.html

PR #880, opened 2018-08-24
Example returns incorrectly

c316:
If no user is logged in then

```
  if (!this.userId) { // evaluates to true
    return true; // so if no user is logged in the example shows editableBy is true
  }
```

TODO:


lorensr:
Thanks @c316 ! 😊

c316:
@lorensr  You're welcome

PR #887, opened 2018-09-03
Removed unneccesary use of simply/easily and the like

kdekooter:


content/code-style.md
```diff
@@ -41,7 +41,7 @@ secondStatement();
 
 <h3 id="automatic-error-checking">Automatic error checking</h3>
 
-Having a consistent style means that it's easier to adopt standard tools for error checking. For example, if you adopt a convention that you must always use `let` or `const` instead of `var`, you can now use a tool to ensure all of your variables are scoped the way you expect. That means you can avoid bugs where variables act in unexpected ways. Also, by enforcing that all variables are declared before use, you can easily catch typos before even running any code!
```
abernix: Removing `easily` here works, but as another option I might have recommended changing it to say "more easily" (e.g. "you can more easily catch typos").

That is to say, as a general rule I think it's acceptable to use _relative_ comparison, so long as we don't call things "easy" without substantiation.

Just to repeat though: I'm not asking for this change, purely pointing out my thoughts; the removal of `easily` works just as well here.

content/code-style.md
```diff
@@ -81,7 +81,7 @@ We recommend choosing and sticking to a JavaScript style guide and enforcing it
 
 We recommend using the [Airbnb eslint configuration](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb) which verifies the Airbnb styleguide.
 
-Below, you can find directions for setting up automatic linting at many different stages of development. In general, you want to run the linter as often as possible, because it's the fastest and easiest way to identify typos and small errors.
+Below, you can find directions for setting up automatic linting at many different stages of development. In general, you want to run the linter as often as possible, because it's a fastest and easy way to identify typos and small errors.
```
abernix: In this new context, `fastest` should become `fast`.  I might still lose the `easy` and only call it "fast". 😉

content/build-tool.md
```diff
@@ -118,7 +118,7 @@ CSS framework compatibility should be a primary concern when picking a pre-proce
 An important feature shared by all of the available CSS pre-processors is the ability to import files. This lets you split your CSS into smaller pieces, and provides a lot of the same benefits that you get from JavaScript modules:
 
 1. You can control the load order of files by encoding dependencies through imports, since the load order of CSS matters.
-2. You can create reusable CSS "modules" that just have variables and mixins and don't actually generate any CSS.
```
abernix: I think a lot of `just` usage which has been removed in this PR had previously been _okay_, though I completely understand where you're coming from.  _Just_ letting you know my thinking. 😉

content/code-style.md
```diff
@@ -118,7 +118,7 @@ For more details, read the [Getting Started](http://eslint.org/docs/user-guide/g
 
 <h3 id="eslint-editor">Integrating with your editor</h3>
 
-Linting is the fastest way to find potential bugs in your code. Running a linter is usually faster than running your app or your unit tests, so it's a good idea to run it all the time. Setting up linting in your editor can seem annoying at first since it will complain often when you save poorly-formatted code, but over time you'll develop the muscle memory to just write well-formatted code in the first place. Here are some directions for setting up ESLint in different editors:
```
abernix: 👍

content/code-style.md
```diff
@@ -135,7 +135,7 @@ A side note for Emmet users: You can use *\<ctrl-e\>* to expand HTML tags in .js
 
 <h4 id="eslint-atom">Atom</h4>
 
-Using ESLint with Atom is simple. Just install these three packages:
```
abernix: 👍

content/code-style.md
```diff
@@ -148,7 +148,7 @@ Then **restart** (or **reload** by pressing Ctrl+Alt+R / Cmd+Opt+R) Atom to acti
 
 <h4 id="eslint-webstorm">WebStorm</h4>
 
-WebStorm provides [these instructions for using ESLint](https://www.jetbrains.com/webstorm/help/eslint.html). After you install the ESLint Node packages and set up your `package.json`, just enable ESLint and click "Apply". You can configure how WebStorm should find your `.eslintrc` file, but on my machine it worked without any changes. It also automatically suggested switching to "JSX Harmony" syntax highlighting.
```
abernix: This is a great example of where removing `just` makes sense.

content/data-loading.md
```diff
@@ -607,7 +607,7 @@ Things can get more complicated; for instance you may want to deal with document
 
 <h3 id="publications-as-rest">Accessing a publication as a REST endpoint</h3>
 
-The opposite scenario occurs when you want to publish data to be consumed by a 3rd party, typically over REST. If the data we want to publish is the same as what we already publish via a publication, then we can use the [simple:rest](https://atmospherejs.com/simple/rest) package to do this really easily.
```
abernix: 👍 "More easily" would have also been acceptable, I think.  Removing it is great.  (and _some_  future version of the docs that had a CI test/linting step which forbid terms like _easy_ would welcome this change.)

content/code-style.md
kdekooter: agreed

content/code-style.md
kdekooter: missed that one, you are right

content/build-tool.md
kdekooter: :-)

PR #892, opened 2018-09-24
Document how to properly handle multiple subscriptions

charpeni:
Add documentation on how to handle properly multiple subscriptions using [`Array.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) and [`Array.every`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every).

I've noticed a useless pattern across many codebases where devs would assign each handle to a variable, then may assign `handle.ready()` to another variable (ex: `const xIsReady = handle.ready()`), then they would check each of them inside a huge condition to compute the ready/loading state.

Using [`Array.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) and [`Array.every`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) simplify everything as we only have to add a `Meteor.subscribe` into the array and let him handle the ready state.


lorensr:
Thanks! ☺️

PR #893, opened 2018-09-24
Prevent exporting Container as global variable

charpeni:
I noticed an unwanted behavior across Meteor codebases that I've been able to track down here.

If we do the following, we do not simply name our container, we create a **global variable** that we export:

```js
export default ListPageContainer = withTracker(({ id }) => {
  // ...
})(ListPage);
```

The syntax below is not valid, we can't `export default` an inline initialized variable. As a side effect, we'll have `ListPageContainer` in our global scope.

We would have to..

Declare the const before using `export default` keyword:

```js
const ListPageContainer = withTracker(({ id }) => {
  // ...
})(ListPage);

export default ListPageContainer;
```

Or if they're in the same file, we can `export default` our base class, and `export` our container as an inline `const`.

```js
export default class ListPage extends Component {
  // ...
}

export const ListPageContainer = withTracker(({ id }) => {
  // ...
})(ListPage);
```

Then, we'll have to import like this:

```js
import ListPage, { ListPageContainer } from "...";
```

cc @abernix A serious unattended behavior that probably worth your eyes on. 👀

### DIFF
--- a/content/react.md
+++ b/content/react.md
@@ -215,7 +215,7 @@ ListPage.propTypes = {
 };
 ```
 
-Then we create a `ListPageContainer` container component which wraps it and provides a data source:
+Then we create a container component which wraps it and provides a data source:
 
 ```js
 import { Meteor } from 'meteor/meteor';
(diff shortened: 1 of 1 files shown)


lorensr:
Thanks for the contribution! Edited to use the suggested method from the get-go ☺️

PR #899, opened 2018-10-15
Update build-tool.md to remove stylus

skirunman:
Stylus is no longer supported in Meteor, and was never that popular, so remove from guide.  https://github.com/meteor/meteor/pull/9445


lorensr:
Heyy, I used stylus with meteor for 2 years! 😄

PR #906, opened 2018-10-23
Vue SSR guide

chris-visser:
I've extended the Vue section of the guide to include how to do SSR. Feel free to modify or extend or maybe bitch about the approach so I can take some extra time to fix that :)


lorensr:
Thanks for writing! ☺️

PR #927, opened 2018-11-19
Fixed invalid markdown markup within Vue SSR guide.

amka:
Fixed non-closed code block that crashes all markup after itself.


lorensr:
Thanks! 😊

PR #928, opened 2018-11-21
Formatting and list rendering

MichaelJCole:
Changed the UL to an OL, hoping it will render correctly.

TODO:


lorensr:
Can check how it will look locally:

https://github.com/meteor/guide/blob/master/CONTRIBUTING.md#running-the-static-site-generator-locally

PR #929, opened 2018-11-21
Add eslint config options

MichaelJCole:
So they don't have to be searched for.

TODO:


SimonSimCity:
Thanks very much for taking the time and drafting out this PR.

Since the guide on another place promotes the usage of pseudo-imports (https://guide.meteor.com/structure.html#importing-meteor-globals) I would rather use them instead of adding any global variables here.

Do you have a reason of advising to disable the `no-console` statement?

The `env` part for `node` or `browser` should (as I think) go into an `overrides` parameter, which is only enables this option inside a `client` or `server` folder to be sure it's not included at any other environment - like following:

```json
{
  "eslintConfig": {
    "overrides": [
      {
        "files": [
          "**/client/**/*.js"
        ],
        "env": {
          "browser": true
        }
      }
    ]
  }
}
```

Btw .. I haven't yet found a reason to set the `env.node` to `true`, but I've found the object `Assets` which (as of the time writing) cannot be imported - at least when I tried it 😢

MichaelJCole:
Hey Simon, thanks for commenting.  I noticed yall were bringing Meteor under new management and that's great.  It's an amazing tech I haven't found a replacement for.  I was keen to get it to work with Vue, but didn't find a reliable solution at the time.

I haven't worked with Meteor since around this PR, so I'd definitely go with what your instincts tell you. The PR may be out of date.

I used console logging alot for debugging server-side, so that's why I no-console.

Re: overrides, I think you are more informed than myself about this.  I vote yes :-)

Thanks for picking this up!

StorytellerCZ:
Updated the PR with feedback from @SimonSimCity

StorytellerCZ:
@MichaelJCole can you please sign the CLA?

StorytellerCZ:
Closing this since the CLA was not signed.

PR #932, opened 2018-11-23
Add explanation on supporting multiple major versions in atmosphere packages

coagmano:
Realised the info from https://github.com/meteor/meteor/blob/devel/History.md#v093-2014-09-25 was missing in the writing atmosphere packages section.


lorensr:
Thanks, sorry for the delay!

PR #946, opened 2018-12-25
Removed dead link

akolybelnikov:
The kadira website does nto provide secure https connection and cannot be open


lorensr:
First want to check whether @arunoda will be fixing old.kadira.io ? ☺️
image

lorensr:
Thanks @akolybelnikov  😊

PR #949, opened 2019-03-07
Formatting and a couple clarifications.

MichaelJCole:
TODO:

### DIFF
--- a/content/vue.md
+++ b/content/vue.md
@@ -300,7 +300,7 @@ You should see:
   - startup times for the server (added to the Time collection on startup)
   - The Meteor settings from your settings file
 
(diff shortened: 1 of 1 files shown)


lorensr:
Thanks ☺️

MichaelJCole:
Hey guys, I'll leave these notes here:

- I think [this section](https://guide.meteor.com/vue.html#vue-and-meteor-realtime-data-layer) is just adding missing code from the previous section.  These could probably be combined.

- I still don't get the [SSR section](https://guide.meteor.com/vue.html#ssr-code-splitting).  The code doesn't fit into the previous examples, and is missing alot of context.  It doesn't make sense to me.

- The last section "Async data - An Interesting Nuxt Feature" has a broken anchor, and I think it just confuses the issue.  Using Meteor with Vue.  I'm a fan of Nuxt.  This section has nothing to do with Meteor integration.

Anyways, happy Thursday!

PR #951, opened 2019-03-27
Note dependency on Xcode 10.1 now that 10.2 is auto-updated by Apple

alanning:
TODO:


alanning:
Hey @lorensr, are you the right person to ping regarding getting this merged?  :-)
Now that Apple is auto-updating everyone's Xcode to 10.2, which is not compatible with latest Meteor, this is something a lot more people will run into.

lorensr:
Hello there Adrian, thanks for the contribution 😄🙏

PR #967, opened 2019-04-28
Update iOS build section now that Xcode 10.2 is supported

alanning:
Merged PR https://github.com/meteor/cordova-plugin-meteor-webapp/pull/78 added support for Xcode 10.2 and is included in webapp v1.7.4 so the iOS section needed to be updated to reflect the recent changes.

TODO:


lorensr:
Thanks 😊

PR #992, opened 2019-07-25
Update suggested Docker images

GeoffreyBooth:
Add https://github.com/disney/meteor-base to the list of suggested Docker images (disclosure: I’m the author); remove two images that are no longer maintained


lorensr:
Thanks! ☺️

PR #1007, opened 2019-09-22
Replace aldeed:simple-schema with npm simpl-schema

vparpoil:
the atmosphere version is only getting critical fixes, has been replaced by the npm version


harryadelb:
https://guide.meteor.com/collections.html#validating-schemas
The validation part could use some modifications; the provided syntax is no longer valid.

The proper syntax is: 
```
Collection._c2._simpleSchema.validate({ foo: 'bar' })
```
Though thinking about it this is somewhat complicated/unneeded, so I guess it's better to remove the part related to validation and just mention that collection2 triggers simpl-schema internally for validations. What do you think guys?

filipenevola:
Hi, please let me know when we have a consensus here so we can merge this PR.

coagmano:
I think these changes are fairly uncontroversial, but I would wait until aldeed/simple-schema-js#344 is fixed since it means the package doesn't work with meteor out of the box without additional configuration

vparpoil:
👋  @filipenevola , it seems that the issue listed above by @coagmano has been fixed !

sebakerckhof:
LGTM

PR #1016, opened 2019-10-23
Replace references to percolatestudio:migrations by npm mgdb-migrator

vparpoil:
Since `mdgb-migrator` has been created from `percolatestudio:migrations` and seems more up to date and maintained, this PR proposes to replace the references to `percolatestudio:migrations` by the npm version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ title: Changelog
 order: 1001
 description: A log of significant changes to the Meteor Guide.
 ---
-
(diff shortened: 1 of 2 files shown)


StorytellerCZ:
While I do like the improvements in that package I dislike the fact that you need to setup a separate connection to the DB through it. It kind of goes against the spirit of Meteor. I think more appropriate option would be to see if we can move `percolate:migrations` to MCP or create a new version there with the improvements.

StorytellerCZ:
I'm closing this issue because we haven't heard anything from the original poster for a while.

PR #1033, opened 2020-04-14
Update method to recompile node modules

sebakerckhof:
This has been possible since 1.8.2 and should now be the preferred method.


lorensr:
👍☺️

PR #1034, opened 2020-04-22
fix link to changelog

alanning:
TODO:


alanning:
The current Contributing Guidelines page has an incorrect link to the changelog doc.  It linked to changelog.md which results in the visitor seeing a "Page Not Found" error message.

This small fix changes the link to point to "changelog.html" which succeeds.

lorensr:
👋 Thanks Adrian 😄

PR #1039, opened 2020-05-19
Update react-native.md

alDuncanson:
Moved meteor out of named imports, as it's a default import.

TODO:


filipenevola:
thanks 😉

PR #1046, opened 2020-08-19
docs(deployment): removes mLab as suggested DBaaS

Addono:
mLab got acquired by MongoDB and their customers are encouraged to migrate to MongoDB Atlas instead. Removing mLab makes sense, as their services are being sunset. MongoDB Atlas is already incorporated in the list of suggested database hosts, so this guides users to the service replacing mLab.

https://docs.mlab.com/mlab-to-atlas/

TODO:


filipenevola:
Thanks @Addono

Issue #1047, opened 2020-09-11
Remove or improve the part about crosswalk


namenotrequired:
I mean this section: https://guide.meteor.com/cordova.html#what-environment -> Android

This section is mostly about Crosswalk. As far as I can tell,

- Android 4.4 and below have very small market share in 2020
- The two links to pages on crosswalk-project.org redirect to an unrelated website which is confusing
- Talking about it as a "plugin" is confusing, because people come to this same article to see how to install cordova plugins. But this "plugin" it is installed as a package. People see the code example (`meteor add crosswalk`) and think that's how you install cordova plugins.

Can we remove this section, or is crosswalk still relevant in 2020?

If it is, can we clarify the confusion?

lorensr:
I'm fine removing

filipenevola:
Yes, please remove @namenotrequired

namenotrequired:
Done: https://github.com/meteor/guide/pull/1049

PR #1049, opened 2020-09-13
Remove crosswalk

namenotrequired:
As discussed here: https://github.com/meteor/guide/issues/1047

I removed the section about problems with web views in Android 4 and previous, and how crosswalk helps you solve them.

For those already using crosswalk, I decided to keep the crosswalk instructions in the "deploying to production" section. I only fixed the link.

TODO


lorensr:
Thanks! ☺️

PR #1053, opened 2020-09-30
Update build-tool.md

enryco:
Updated TypesScript instructions/libs and added `tsconfig` example from _meteor create --typescript_
Details:
- Don't remove ECMAScript, this will break meteor
- Add the official typescript library


lorensr:
Thanks! Is that just an example file/values? Are certain properties required?

enryco:
This is the configuration and the tsconfig file which you will get if you run `meteor create --typescript`. I think that's a good starting point for existing meteor projects as well (which was my case)

PR #1061, opened 2020-11-03
Add note about rate limiting and DDP mixin

harryadel:
Added [ddp-rate-limiter-mixin](https://github.com/nlhuykhang/ddp-rate-limiter-mixin) as suggested by @aliogaili and another note about DDP rate limiting necessity.


lorensr:
Thanks ☺️

PR #1067, opened 2021-01-17
Remove obsolete section with dynamic imports

realyze:
Typescript now supports dynamic imports—and has for almost three years—so this section (which I added four years ago :) ) is misleading and should be removed.

### DIFF
--- a/content/build-tool.md
+++ b/content/build-tool.md
@@ -113,12 +113,6 @@ If you want to add TypeScript from the point of project creation, as of Meteor 1
 meteor create --typescript name-of-my-new-typescript-app
 ```
 
-<h4 id="typescript-conditional-imports">Conditional imports</h4>
-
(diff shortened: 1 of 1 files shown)


filipenevola:
Hi @realyze nested imports are not the same as dynamic imports.

Nested imports is a unique feature of Meteor.

Read more about nested imports here https://zodern.me/posts/nested-imports/

realyze:
@filipenevola Ah got it, thanks! I still think this section is misleading now that dynamic `import` is a thing in TypeScript - it's definitely very strongly preferred to use dynamic `import` that `require`, wouldn't you agree?

filipenevola:
Hi @realyze, they are different things and they solve different problems, in short:
`nested`: defer the evaluation of modules, good to speed up startup without removing code from initial bundle (in case you know you will need then very soon)
`dynamic`: defer the download (availability) of modules

So both can and should be used depending on each case.

Meteor with plain JS support both. In Typescript, as the language does not support `nested`, it only supports `dynamic`

PR #1071, opened 2021-02-07
Fix typo in react-router description

edemaine:
Fix a small plurality error describing react-router. Also add a(nother) link to react-router for easier reading.


lorensr:
Thanks ☺️

PR #1076, opened 2021-03-25
security/helmet provide production CSP example and fix #1075

### LINKED ISSUE #1075: Helmet Example Link is now a 404
In the [CSP guide for helmet](https://guide.meteor.com/security.html#csp) there was an example linked to a gist. However it is 404 now: https://gist.github.com/dhrubins/31248506873a006ba5e6b459a81855c2

Fortunately I have used the Gist as basis for our setup and extended it. Due to HMR there is also now an additional entry to make in order to work with `sha26-<hash>` pattern for script src.

I would propose to provide an out-of-the box example that passes CSP validators to have "built-in" security for devs that are unfamiliar with CSP.

I will wrap up and comment, once the CSP is generic enough.

jankapunkt:
FIxes #1075 with cover for 2.0 and < 2.0 specifics

### DIFF
--- a/content/security.md
+++ b/content/security.md
@@ -515,7 +515,162 @@ WebApp.connectHandlers.use(
 
 Helmet supports a large number of directives, users should further customise their CSP based on their needs. For more detail please read the following guide: [Content Security Policy](https://helmetjs.github.io/docs/csp/). CSP can be complex, so in addition there are some excellent tools out there to help, including [Google's CSP Evaluator](https://csp-evaluator.withgoogle.com/), [Report-URI's CSP Builder](https://report-uri.com/home/generate), [CSP documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src) from Mozilla and [CSPValidator](https://cspvalidator.org/).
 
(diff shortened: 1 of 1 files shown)


content/security.md
```diff
@@ -515,7 +515,162 @@ WebApp.connectHandlers.use(
...
+const data = 'data:'
+const unsafeEval = '\'unsafe-eval\''
+const unsafeInline = '\'unsafe-inline\''
+
+/**
+ * Creates a helmet-compatible CSP-configuration
+ * @param externalHostUrls {Array|undefined} Optional array with allowed hosts
+ * @return {Object} a CSP configuration object
+ */
+export function createCSPOptions (externalHostUrls = []) {
```
StorytellerCZ, line 539: Would be better to include here the Helmet implementation than just a function to export.

StorytellerCZ (changes requested):

content/security.md
jankapunkt: You are right, I am so used to separate these things from each other :-D

jankapunkt: I removed the functions and rewrote the code in a more linear fashion to make it much more understandable and finally call helmet at the end with the created options.

StorytellerCZ (approved):

PR #1081, opened 2021-08-16
2.4 migration guide

StorytellerCZ:
In preparations for 2.4 release here is a quick migration guide as they were done until v1.7. I will look into creating ones for the missing versions as well to have guide complete in the future (maybe in this PR if I can manage it before it gets merged, but they will have to be backported to the old versions).

### DIFF
--- a/content/1.10-migration.md
+++ b/content/1.10-migration.md
@@ -1,0 +1,64 @@
+---
+title: Migrating to Meteor 1.10
+description: How to migrate your application to Meteor 1.10.
+---
+
+Most of the new features in Meteor 1.10 are either applied directly behind the scenes (in a backwards compatible manner) or are opt-in. For a complete breakdown of the changes, please refer to the [changelog](http://docs.meteor.com/changelog.html).
+
+The above being said, some breaking changes to note and migration steps for a bug that you might encounter.
+
+<h3 id="mongo-exit-62">Unexpected mongo exit code 62</h3>
+
+If you get `Unexpected mongo exit code 62. Restarting.` when starting your local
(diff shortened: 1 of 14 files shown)


filipenevola (changes requested):
Why do we need to have this here?

Before Meteor 1.8 we didn't have migration steps as a section in the History.md but now we do so we should centralize this there.

StorytellerCZ:
@filipenevola ease of access, especially for users that are upgrading from older releases (easier to have this extract) or new users who are still unfamiliar with the changelog or visiting the repository to get news.

filipenevola:
> @filipenevola ease of access, especially for users that are upgrading from older releases (easier to have this extract) or new users who are still unfamiliar with the changelog or visiting the repository to get news.
> 
> But it is a worthy discussion if this is something that we should have or remove.

The changelog is in the [docs](https://docs.meteor.com/changelog.html), it's enough. I don't see how spreading the content here would help.

Ideally we should even remove History.md and start using Github Releases as almost every project does.

StorytellerCZ:
We could do the releases. I can start on that if you want and then I can remove migrations from here entirely.

filipenevola:
We can close this cycle like this and decide what do to next for Meteor 2.5.

StorytellerCZ:
@filipenevola I'm not sure I understand, do you want to merge this or leave this for 2.5?

filipenevola:
> @filipenevola I'm not sure I understand, do you want to merge this or leave this for 2.5?

Now that this work in already done you can merge it and we discuss how we want to move forward starting from 2.5

